PR #8, opened 2017-05-30
enable codecov.io

zaibon:


test.sh
```diff
@@ -3,7 +3,7 @@ set -e
 pushd store
     echo "Build store"
     go get ./...
-    go test ./...
+    bash codecov.sh
```
Hamdy: can u please add to the readme a section for testing pre-requisites?
As testing needs nodejs package to be installed (nodejs includes npm)
Also raml2html package is required

store/codecov.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+echo "" > coverage.txt
+
+for d in $(go list ./... | grep -v vendor); do
+    go test -race -coverprofile=profile.out -covermode=atomic $d
+    if [ -f profile.out ]; then
+        cat profile.out >> coverage.txt
```
Hamdy: I've tried the tests but coverage.txt seems to be empty.
Is this OK?

zaibon: It's normal since we don't have any tests yet. but we'll have to change that and raise the test coverage as we go

test.sh
zaibon: in travis I also make sure that the raml is valid and that we can generate both the server/client code and the html documentation. This should not impact on the test of the code itself.
For the code itself there is not prerequis. we use standard go tools

PR #18, opened 2017-06-13
Introducing namespace statistics

Hamdy:


store/NamespaceStat.go
```diff
@@ -13,3 +15,35 @@ func (s NamespaceStat) Validate() error {
...
 }
+
+type Stat struct{
+	NamespaceStat
+	NrRequests int64
+	creationDate time.Time
+}
+
+
+func (s *Stat) toBytes() []byte{
```
zaibon: instead of `toBytes` and `fromBytes` why don't use implement the `io.Reader` and `io.Writer` interface. This is more idiomatic and it will make easy to plug this with other method of the go lib

store/namespaces_api.go
```diff
@@ -5,3 +5,11 @@ type NamespacesAPI struct {
 	db     *Badger
 	config *settings
 }
+
+func (api NamespacesAPI) getDb() *Badger{
```
zaibon: Don't prepend with `get` just call is `DB` and `Config`

store/namespaces_api_Createnamespace.go
```diff
@@ -55,7 +57,26 @@ func (api NamespacesAPI) Createnamespace(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	respBody := &Namespace{
+	// Add stats
+	// We assume namespaces and object names can't contain (_)
```
zaibon: don't assume, enforce this. 
In the raml specify the pattern of a valid object ID and make sure we validate this when creating object.

store/namespaces_api_Createobject.go
```diff
@@ -14,22 +14,7 @@ import (
...
-		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
-		return
-	}
-
-	// NOT FOUND
-	if v == nil {
-		http.Error(w, "Namespace doesn't exist", http.StatusNotFound)
-		return
-	}
+	nsid := mux.Vars(r)["nsid"]
```
zaibon: Now that you have the middleware that check the existence of the namespace. why don't you get the namespace from the context ?

store/namespaces_if.go
```diff
@@ -62,25 +62,110 @@ type NamespacesInterface interface { // nsidaclPost is the handler for POST /nam
 	// Createnamespace is the handler for POST /namespaces
 	// Create a new namespace
 	Createnamespace(http.ResponseWriter, *http.Request)
+
+	// Get db object
+	getDb() *Badger
```
zaibon: This doesn't need to be in the interface

store/namespaces_if.go
```diff
@@ -62,25 +62,110 @@ type NamespacesInterface interface { // nsidaclPost is the handler for POST /nam
...
+
+	// Get Settings object
+	getConfig() *settings
```
zaibon: idem

store/namespaces_if.go
Hamdy: can u explain how? I get compilation error if I removed that from interface

store/namespaces_api_Createobject.go
Hamdy: here' am just getting the nsid (key)
we don't need the namespace object here in this method

PR #20, opened 2017-06-28
Reservations

Hamdy:


specs/raml/sdstor.raml
```diff
@@ -152,6 +165,35 @@ types:
   post:
     body:
       type: NamespaceCreate
+
+  /stats:
+     securedBy: [ itsyouonline : { scopes: ["user:name"] } ]
+     post:
```
zaibon: Whats the use case of this ?
The store statistics is a read-only information for the user. It should not be allow to modify it.

store/ACLEntry.go
```diff
@@ -16,3 +16,68 @@ func (s ACLEntry) Validate() error {
...
 }
+
+/*
+
+| Admin  | Read  | Write  | Delete  |
+|--------|-------|--------|---------|
+| 1 byte | 1 byte| 1 byte | 1 byte  |
+
+ */
+func(s ACLEntry) ToBytes() []byte{
```
zaibon: bit mask would probably be faster

store/NamespaceCreate.go
```diff
@@ -12,3 +13,126 @@ type NamespaceCreate struct {
...
+	return s.Save(db, config)
+
+}
+
+func (s NamespaceCreate) Exists(db *Badger, config *settings) (bool, error){
+	return db.Exists(s.Label)
+}
+
+func (s NamespaceCreate) Save(db *Badger, config *settings) error{
+	return db.Set(s.Label, s.ToBytes())
```
zaibon: Shouldn't we also have a prefix for Namespace ?
So we can list them really fast ?
Cause now you have to walk over the all database to list the namespaces

store/main.go
```diff
@@ -110,6 +131,28 @@ func main() {
 			Value:       100,
 			Destination: &settings.Iterator.PreFetchSize,
 		},
+
+		cli.StringFlag{
```
zaibon: I don't think we should make these prefix configurable.
This is internal business logic.

store/namespaces_api_ListReservations.go
```diff
@@ -2,9 +2,101 @@ package main
...
+	opt.PrefetchSize = api.config.Iterator.PreFetchSize
+
+	it := api.db.store.NewIterator(opt)
+	defer it.Close()
+
+	startingIndex := (page-1)*per_page + 1
+	counter := 0 // Number of objects encountered
+	resultsCount := per_page
+
+	for it.Rewind(); it.Valid(); it.Next() {
```
zaibon: Use `it.Seek(prefix)` and stop iteration when the prefix is not present anymore.

zaibon: See this commit: they added a new method on the iterator object to do exactly this:
https://github.com/dgraph-io/badger/commit/eaad7ebcf7437492bdf65ce674976a0fd531e71e

specs/raml/sdstor.raml
Hamdy: Mistake, this should have been a separate section
so basically there's a store stats which holds the available total size that store can handle and we can change that limit in runtime through a rest call.
It holds also total amount of space used for all reservations. 
this is efficient and enables us to determine whether a new reservation is applicable or not without going through all reservation records in db and get the total amount of space used for valid ones in runtime when creating new reservations.

store/ACLEntry.go
Hamdy: can u explain how I can do bit masking in this case?

store/main.go
Hamdy: OK

store/NamespaceCreate.go
Hamdy: Listing namespaces is not a frequent operation. it's meant for admin of the 0-stor not the normal users.

store/namespaces_api_ListReservations.go
Hamdy: OK

store/NamespaceCreate.go
zaibon: Doesn't matter if it's called a lot or not, walking the full database is bad

specs/raml/sdstor.raml
zaibon: >  and we can change that limit in runtime through a rest call.

This is not a requirement at the moment, and + if we do that, then it needs to be protected with a jwt that confirm we are the admin of the store, so ` [ itsyouonline : { scopes: ["user:name"] } ]` is wrong.

>this is efficient and enables us to determine whether a new reservation is applicable or not without going through all reservation records in db and get the total amount of space used for valid ones in runtime when creating new reservations.

This is great but this should be some internal knowledge that the store need for himself. I don't get why it needs to be exposed.

store/ACLEntry.go
zaibon: https://play.golang.org/p/Dm3X-akr31

zaibon:
Merging this cause quite some logic is part of this PR and it would be hard to split it now.
but some refactoring on the code is needed and going to be provided in next PR

PR #37, opened 2017-07-17
Unittests

Hamdy:


store/rest/models/object.go
```diff
@@ -5,55 +5,76 @@ import (
...
+	size := 10 + 1 + idSize + nsSize + plSize + tagsSize + crcSize
 
 	result := make([]byte, size)
 	binary.LittleEndian.PutUint16(result[0:2], uint16(idSize))
 	binary.LittleEndian.PutUint16(result[2:4], uint16(nsSize))
-	binary.LittleEndian.PutUint16(result[4:6], uint16(payloadSize))
-	binary.LittleEndian.PutUint16(result[6:8], uint16(tagSize))
+	binary.LittleEndian.PutUint16(result[4:6], uint16(crcSize))
+	binary.LittleEndian.PutUint16(result[6:8], uint16(plSize))
+	binary.LittleEndian.PutUint16(result[8:10], uint16(tagsSize))
```
iwanbk: Maybe just a matter of taste, but i think we can do
```
buf := new(bytes.Buffer)
......
binary.Write(buf, binary.LittleEndian, uint16(crcSize)
binary.Write(buf, binary.LittleEndian, uint16(plSize)
binary.Write(buf, binary.LittleEndian, uint16(tagSize)
result := buf.Bytes()
```

zaibon: Looks better indeed, will create a task just for that

PR #40, opened 2017-07-18
update JWT creation

### LINKED ISSUE #33: Review reservation logic
The current implementation of the reservations and how the token are generated is wrong. 
This parts needs to be re-written.

Concepts:
At the creation of a reservation, 2 token are created
- ReservationToken : This is used to managed the reservation. it's needed if user want to renew a reservation
- DataAccessToken: This token is needed to access the reservation and read/write object on the store.
Multiple DataAccessToken can be created for different users.

Both of these token are JWT, the current implementation doesn't generate JWT.

Requirements:
- the store needs to generate a key pair that is going to b used to sign the JWT.
- Token generation needs to be updated to generate JWT

zaibon:
Create real JWT token for reservation and data access token

fixes #33


scripts/generate_jwt_key.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"crypto/rand"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+)
+
+var keySize = flag.Int("size", 2048, "size of the key")
+var out = flag.String("out", "", "destination of the generate key. If empty output on stdout")
```
iwanbk:
```go
var (
   keySize..
    out...
)
```
Make it more consistent

store/jwt/data_access.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"errors"
+	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/zero-os/0-stor/store/rest/models"
+)
+
+var ErrWrongUser = errors.New("data access token not valid, wrong user")
+var ErrWrongACL = errors.New("data access token not valid, wrong permission")
```
iwanbk: same as previous comment

store/jwt/data_access.go
```diff
@@ -0,0 +1,72 @@
...
+func ValidateDataAccessToken(tokenString, user, namespace string, acl models.ACLEntry, key []byte) error {
+	token, err := jwt.ParseWithClaims(tokenString, &dataAccessClaims{}, func(token *jwt.Token) (interface{}, error) {
+		return key, nil
+	})
+
+	if alg, present := token.Header["alg"]; !present || alg != jwt.SigningMethodHS256.Alg() {
+		return ErrWrongAlg
+	}
+
+	if claims, ok := token.Claims.(*dataAccessClaims); ok && token.Valid {
```
iwanbk: why not
```go
if claims, ok := token.Claims.(*dataAccessClaims); ok && !token.Valid  {
     return ...
}
.....// rest of the code
```

store/jwt/data_access.go
```diff
@@ -0,0 +1,72 @@
...
+		if (acl.Admin && !claims.ACL.Admin) ||
+			(acl.Read && !claims.ACL.Read) ||
+			(acl.Write && !claims.ACL.Write) ||
+			(acl.Delete && !claims.ACL.Delete) {
+			return ErrWrongACL
+		}
+		return nil
+	}
+
+	return err
```
iwanbk: It seems the `err` here will always be `nil`

store/jwt/reservation.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	return token.SignedString(key)
+}
+
+func ValidateReservationToken(tokenString, namespace string, key []byte) (string, error) {
+	token, err := jwt.ParseWithClaims(tokenString, &reservationClaims{}, func(token *jwt.Token) (interface{}, error) {
+		return key, nil
+	})
```
iwanbk: need to check
```go
if err != nil
```

store/rest/actions_namspace.go
```diff
@@ -392,22 +389,19 @@ func (api NamespacesAPI) nsidaclPost(w http.ResponseWriter, r *http.Request) {
...
+		dataToken, err = jwt.GenerateDataAccessToken(reqBody.Id, reservation, reqBody.Acl, api.jwtKey)
 		if err != nil {
 			log.Errorln(err.Error())
 			http.Error(w, "Internal Server Error", http.StatusInternalServerError)
 			return
 		}
 	}
 
-
 	w.Header().Set("Content-Type", "application/json")
```
iwanbk: It seems `go-raml` really need to generate this line

store/rest/actions_object.go
```diff
@@ -121,7 +119,7 @@ func (api NamespacesAPI) Createobject(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		// Check files are identical, otherwise throw conflict
-		if oldFile.CRC != file.CRC{
+		if oldFile.CRC != file.CRC {
```
iwanbk: I don't think `CRC` can be used to check for duplicate content. It is only to check for data corruption

store/rest/middleware/namespace_stat.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+
+			err = nsStats.Decode(b)
+			if err != nil {
+				log.Errorln(err.Error())
+				return
+			}
+
+			nsStats.NrRequests++
+			if err != nil {
```
iwanbk: why we need this?

iwanbk:
LGTM

PR #47, opened 2017-07-19
WIP: create manager for the different types of objects

zaibon:
The idea is to have an interface that deal with the common operation on
all types

Later on, the goal is to be able to re-use these manager from the grpc
interface so we only have ony place for the buisness logic of each
objects


store/rest/namespace_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/zero-os/0-stor/store/db"
+	"github.com/zero-os/0-stor/store/manager"
+	"github.com/zero-os/0-stor/store/rest/models"
+)
+
+// ErrNamespaceConflict is the error return when trying to create a
+// namespace with a label that alredy exists
+var ErrNamespaceConflict = errors.New("Namespace with this label alresdy exists")
+
+var _ (manager.NamespaceManager) = (*namespaceManager)(nil)
```
iwanbk: what this line does?

store/rest/namespace_manager.go
```diff
@@ -0,0 +1,130 @@
...
+		db: db,
+	}
+}
+
+func (mgr *namespaceManager) Create(label string) error {
+	namespace := models.Namespace{
+		NamespaceCreate: models.NamespaceCreate{Label: label},
+	}
+
+	exists, err := mgr.db.Exists(namespace.Key())
```
iwanbk: i think once the code reach the db, it could be shared between Rest & grpc, no?

store/rest/namespace_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return mgr.createNamespaceStats(namespace.Label)
+}
+
+// List return a list of namespaces
+// start and count are used for the pagination
+// if count < 0, then it returns everything
+func (mgr *namespaceManager) List(start, count int) ([]*models.Namespace, error) {
+	list, err := mgr.db.Filter(models.NAMESPACE_PREFIX, start, count)
```
iwanbk: same as previous comment

zaibon:
> But i think the code that deal with DB could be shared between grpc & rest and shoukd not exist in the manager

Indeed, the DB interface has been made to deal only with `[]byte` so it should be agnostic of the layer above.

store/rest/namespace_manager.go
zaibon: This is a small trick I use to validate that a type implement an interface. With this line I have a compilation error if I miss a method or if a signature is not valid.

store/rest/namespace_manager.go
zaibon: How should we modify the code to achieve this ?

iwanbk:
Oh, i think i might misunderstood your idea.

I have two interpretations:

- from your idea, it is an `interface`, i thought each `grpc` and `rest` should implement this interface. But your WIP code doesn't reflect this

- from your code, it seems your `interface` will be called by both `grpc` and `rest` handler. So your WIP already good, no need to rework the DB part. It doesn't necessarily to be an interface though. But being an interface will help in unit test.

store/rest/namespace_manager.go
iwanbk: i think i misunderstood your idea, see https://github.com/zero-os/0-stor/pull/47#issuecomment-316422710

zaibon:
Yep indeed, that's the second interpretation. 
I would like to be able to re-use the managers in both grpc and rest. So we don't have twice the same bushiness logic.

iwanbk:
> from your code, it seems your interface will be called by both grpc and rest handler. So your WIP already good, no need to rework the DB part. It doesn't necessarily to be an interface though. But being an interface will help in unit test.

and i prefer this idea

So, both grpc & rest handler will be like this:

```go
func theHandler() {
     someDecoding()
     someInputValidation()
     callingTheCommonInterface()
     sendTheResponse()
}
```

zaibon:
Yep Exactly

PR #65, opened 2017-07-27
Refactoring

Hamdy:


iwanbk:
quick comment:
- delete `client` dir
- move `goraml` dir to server

iwanbk:
I think this dir should be deleted https://github.com/zero-os/0-stor/tree/refactoring/server/.db, no?
and add it to `.gitignore`

PR #93, opened 2017-07-31
Client refactor

### LINKED ISSUE #90: Write some example for each component and a example that uses multiple component
<a href="https://github.com/zaibon"></img></a> **Issue by [zaibon](https://github.com/zaibon)**
_Wednesday Jul 19, 2017 at 16:40 GMT_
_Originally opened as https://github.com/zero-os/0-stor-lib/issues/21_

----

### LINKED ISSUE #92: 0-stor client
<a href="https://github.com/iwanbk"></img></a> **Issue by [iwanbk](https://github.com/iwanbk)**
_Monday Jul 24, 2017 at 04:54 GMT_
_Originally opened as https://github.com/zero-os/0-stor-lib/issues/24_

----

0-stor client implementation
- [x] rest client (#22)
- [x] grpc client (not in current milestone)
- [x] IYO helper
- [x] client writer pipe
- [x] client reader pipe (pipe.ReadPipe)

0-stor client is a pipe of writers (implementing io.Writer interface) and readers (io.Reader)  as described at  the [requirements](https://github.com/zero-os/0-stor-lib/blob/master/specs/requirements.md). 

The usage of the client follows the pseudo language described at https://github.com/zero-os/0-stor-lib/blob/master/specs/principles.md#pseudo-language-to-use-lib-example-in-python. Instead of using command/func to do the configuration, it uses configuration file as described at #16. Configuration using command/func can be added later.

The IYO helper is small package which provide API to interact with itsyou.online. It wraps the official itsyou.online library and make it easier to use.

It must support both grpc & REST but only one at a time.

iwanbk:
Fixes #90 and Fixes #92 

- fully move client code to this repo
- all libs has consistent API now
  It is not using io.Writer and io.Reader like the previous person. It now use block Writer/Reader interface because most of the libs works in block instead of stream.
- add pipe example https://github.com/zero-os/0-stor/tree/client-refactor/client/pipe
- tested using to store file to local 0-stor https://github.com/zero-os/0-stor/tree/client-refactor/client/cmd/cli


iwanbk:
The meta package is only quick & simple solution which doesn't follow the spec.
It will be finished in next milestone (milestone-7)

PR #99, opened 2017-08-01
Iyo

Hamdy:


server/test/rest_test.go
```diff
@@ -33,5 +34,15 @@ func getTestAPI(t *testing.T) (string, db.DB, func()) {
 		srv.Close()
 	}
 
-	return srv.URL, db, clean
+	iyoClient := itsyouonline.NewClient("0stor", "b_jZHOchZ226a3iWgFu4iOWe2hRE", "3-Pv44izu2Ge0IXeXhBIdN34R_-w")
```
zaibon: App ID and secret should not be hard-coded. Instead, check the environment variable and extract app ID and secret from there. So in travis we can configure these environment variable and not leak some info in the repo.

server/api/rest/namespacesnsid_if.go
```diff
@@ -47,37 +46,14 @@ type NamespacesInterface interface { // UpdateReferenceList is the handler for P
...
-	 //r.Handle("/namespaces/{nsid}/reservations/{id}", alice.New(NewCreateNamespaceMiddleware(badger.Database).Handler).Then(http.HandlerFunc(i.reservationsidGet))).Methods("GET")
-	 //r.Handle("/namespaces/{nsid}/reservations/{id}", alice.New(NewCreateNamespaceMiddleware(badger.Database).Handler).Then(http.HandlerFunc(i.UpdateReservation))).Methods("PUT")
-	 //r.Handle("/namespaces/{nsid}/reservations", alice.New(NewCreateNamespaceMiddleware(badger.Database).Handler).Then(http.HandlerFunc(i.ListReservations))).Methods("GET")
-	 //r.Handle("/namespaces/{nsid}/reservations", alice.New(NewCreateNamespaceMiddleware(badger.Database).Handler).Then(http.HandlerFunc(i.CreateReservation))).Methods("POST")
-	 //r.Handle("/namespaces/{nsid}", alice.New(NewCreateNamespaceMiddleware(badger.Database).Handler).Then(http.HandlerFunc(i.GetNameSpace))).Methods("GET")
-
-	}
+	 r.Handle("/namespaces/{nsid}/objects/{id}/references", alice.New(NewOauth2itsyouonlineMiddleware().Handler).Then(http.HandlerFunc(i.UpdateReferenceList))).Methods("PUT")
```
zaibon: Here you can instanciate the middleware only once, and pass it to all routes

server/api/rest/oauth2_itsyouonline_middleware.go
```diff
@@ -35,29 +35,15 @@ MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAES5X8XrfKdx9gYayFITc89wad4usrk0n2
...
-		s = append(s, scope)
-	}
-
+func NewOauth2itsyouonlineMiddleware() *Oauth2itsyouonlineMiddleware {
 	om := Oauth2itsyouonlineMiddleware{
-		scopes: s,
+		describedBy:"headers",
+		field: "Authorization",
+		url: oauth2ServerUrl,
+		db: badger.Database,
```
zaibon: I would prefer that we pass the db.DB as an argument. 
Also, why do we have this global `badger.Database` ? Global variable are nearly always bad.

Hamdy: I use global db object to avoid altering auto generated code of go-raml
We had before, a function to return db but you removed it after refactoring. so I found another way by using global db object.
As far as I can tell. I don't see a problem in this case by using global object, since it's not an object we alter its state. besides, even the old approach was sharing same object across all requests which is the same as I do currently by using global object

zaibon: I see. 
But still I think the way it's implement in the badger package is wrong.
So now, when you do a `badger.New` it returns you an instance of the `badgerDB`, but also set the global variable in the package. This is confusing and not really nice design for the package.

Or we need to have the global variable in the `db` package and provide method to initialize this gloabal varibale and we use this everywhere. 
Or we remove the global variable altogether.

I would prefer second option so we keep separation of concern and keeps the `db` and `badger` package clean

PR #102, opened 2017-08-02
client Refactor and improve docs:

### LINKED ISSUE #104: add `key` as  block.Writer.WriteBlock argument
Current `WriteBlock(data []byte) WriteResponse` interface make it not possible to be implemented by rest & grpc client, thus the grpc & rest client can't be piped.

We can add `key` in the argument to overcome this issue.

Although the `key` won't be used by some components like compress & encrypt, it still make sense to be added because it means to be used in 0-stor `key, value` context.

With this addition, we might also change the return value from `WriteResponse` which is uggly to more standard `(written int, err error)` return value.

So, the new interface will be 
```
WriteBlock(key, val []byte) (written int, err error)
```

### LINKED ISSUE #105: complete the client.stor.Client interface
current interface is very minimal.

Need to create full interface which reflect all 0-stor server API

iwanbk:
- add more docs about client usage
- improve client API, make it more modular
- Fixes #104 
- Fixes #105
- exposes the 0-stor rest/grpc client interface
- implements `block.Writer` and `block.Reader` interface to 0-stor rest/grpc client, so it can be used in the pipe


client/lib/distribution/stor_distribution.go
```diff
@@ -68,7 +68,11 @@ func (sd StorDistributor) Write(data []byte) (int, error) {
 	return len(data), nil
 }
 
-// WriteBlock implements block.Writer interface
+// WriteBlock implements block.Writer interface.
+// It returns this information in the meta:
+// - Key = 0-stor key
+// - Size = length of the data
+// - Shardas = 0-stor shards
```
zaibon: Type `Shardas`  instead of `Shards`

iwanbk: Fixed

Issue #107, opened 2017-08-03
raml/spec : no response specified for get & list reservation


iwanbk:
get reservation 

![get_reservation](https://user-images.githubusercontent.com/977552/28904890-2e9465a2-7838-11e7-8eed-54352cdc0be9.png)

list reservation 

![list_reservation](https://user-images.githubusercontent.com/977552/28904897-36352292-7838-11e7-9b26-82b21b4942e5.png)

PR #112, opened 2017-08-04
Completes the 0-stor rest client implementation

iwanbk:
Part of #111 

- except methods blocked by #107
- ObjectExist : it doesn't exist in the goraml generated code


client/stor/rest/client.go
```diff
@@ -59,23 +67,52 @@ func (c *Client) ReservationList() ([]common.Reservation, error) {
...
+func (c *Client) ReservationCreate(size, period int64) (reserv *common.Reservation, reservToken string, dataToken string, err error) {
+	req := reservation.ReservationRequest{
+		Period: period,
+		Size:   size,
+	}
+
+	// call func
+	respObj, resp, err := c.client.Namespaces.CreateReservation(c.nsid, req, nil, nil)
+
+	err = checkRespCode200AndErr(err, resp)
```
iwanbk: yes, that func check if return code is 200 >= and < 300, i may need to rename it to `checkRespCode2XXAndErr`

client/stor/rest/client.go
```diff
@@ -114,7 +149,8 @@ func (c *Client) ObjectGet(id []byte) (*common.Object, error) {
 
 // ObjectDelete delete object from the store
 func (c *Client) ObjectDelete(id []byte) error {
-	return nil
+	resp, err := c.client.Namespaces.DeleteObject(encodeID(id), c.nsid, nil, nil)
+	return checkRespCode200AndErr(err, resp)
```
iwanbk: yes, that func check if return code is 200 >= and < 300, i may need to rename it to checkRespCode2XXAndErr

iwanbk: renaming is done

PR #118, opened 2017-08-04
Chunk in pipe & distribution improvements

### LINKED ISSUE #115: client : include chunker to the pipeline
It will become special case, because the read pipe which normally start from the end (from writer point of view) need to be aware of this chunker which exist in the start of write pipe

iwanbk:
Fixes #115 and distribution improvements
@zaibon i also updated the config as we discussed before
- iyo credentials at root config
- shards at root config

distribution improvements:
- do the upload / download concurrently
- restorer is properly using shards in metadata
- restorer only failed if number of failed servers is more than number of parity (M)

update (2017-08-08):
- distribution now writes to and read from metadata server directly (655fb31d7c7bb9fbbc9fe5e670a58c36f75e9c5e : see the reasoning here)
- add `ShardsClient` as wrapper for `stor.Client` which works on multiple shards
- add `stor.Client` to the end of pipe if there is no components which upload to 0-stor server

Tested to upload - download 273M of lede image


iwanbk:
@zaibon i also updated the config as we discussed before
- iyo credentials at root config
- shards at root config

client/config/config.go
```diff
@@ -36,14 +34,19 @@ var (
...
-	StorClient   stor.Config `yaml:"stor_client,omitempty"`
-	MetaShards   []string    `yaml:"meta_shards"`
-	Pipes        []Pipe      `yaml:"pipes" validate:"nonzero"`
+	Organization string `yaml:"organization" validate:"nonzero"`
+	Namespace    string `yaml:"namespace" validate:"nonzero"`
+	Protocol     string `yaml:"protocol" validate:"nonzero"` // rest or grpc
+
+	IYOAppID       string `yaml:"iyo_app_id"`
+	IYOSecret      string `yaml:"iyo_app_secret"`
+	StorPermWrite  bool   `yaml:"stor_perm_write"`  // 0-stor write permission
```
zaibon: What are these used for ?

iwanbk: the idea was to specify the read/write/delete permission, but then turned out it is not good idea and not used anymore in distribution (which created the 0-stor rest client).
It seems better to remove it, right?

iwanbk: removed in latest commit

PR #123, opened 2017-08-07
improve cli

zaibon:
use github.com/urfave/cli
add commands regarding namespaces
- create, delete, set-acl, get-acl


iwanbk:
@zaibon the readme need to be updated

PR #125, opened 2017-08-08
update client cli

Hamdy:


client/cmd/cli/config.yaml
```diff
@@ -23,12 +23,12 @@ pipes:
           data: 1
           parity: 1
           shards:                                             # the address(es) of 0-stor server
-            - http://127.0.0.1:12345
-            - http://127.0.0.1:12346
+            - http://127.0.0.1:8080
+            - http://127.0.0.1:8080
```
iwanbk: I don't think that shards of  same servers will work properly

iwanbk:
```
 it still need the readme update
```
how to use the `cli` etc, as i think the command line option has been changed

PR #132, opened 2017-08-08
Documentation

Hamdy:


client/cmd/cli/config2.yaml
```diff
@@ -0,0 +1,38 @@
+organization: mordor                                # itsyou.online organization name
```
zaibon: remove this file

client/lib/compress/README.md
```diff
@@ -1,16 +1,14 @@
-# Compress
+## Compression
```
zaibon: package is call compress not compression

docs/config.md
```diff
@@ -0,0 +1,59 @@
+## config
```
zaibon: This should be put into a server folder or something. Now it's just at the root of the docs folder which doesn't indicate what this config is about

docs/gettingstarted.md
```diff
@@ -0,0 +1,12 @@
+## Installation
+
+```
+go get -u github.com/zero-os/0-stor
```
zaibon: pretty sue this doesn't do what you expect.

PR #137, opened 2017-08-09
JWT authentification for grpc interface

zaibon:


server/api/grpc/jwt.go
```diff
@@ -0,0 +1,92 @@
...
+	MethodAdmin  Method = "admin"
+)
+
+//ErrNoJWTToken is return when the grpc request doesnt provide a valid jwt token
+var ErrNoJWTToken = errors.New("No jwt token in context")
+var ErrWrongScopes = errors.New("JWT token doesn't contains required scopes")
+
+func validateJWT(ctx context.Context, method Method, label string) error {
+	isTesting, ok := os.LookupEnv("STOR_TESTING")
+	if ok && isTesting == "true" {
```
iwanbk: then this `func` can't be tested?

zaibon: That's the opposite actually. In the unit test of the server/api/grpc I set this variable (https://github.com/zero-os/0-stor/pull/137/files#diff-186a3aef6f4dba95d9a717bed9838c07R9)
So I only test the grpc handler without the need and hassle of dealing with jwt.

But for the full stack test in server/test, there I don't set the variable so I can test the jwt validation works properly

PR #140, opened 2017-08-09
Multipart file upload

### LINKED ISSUE #63: JSON is not good format to send binary data
0-stor is object store, means that it must support storing binary data.

JSON is bad to send binary data because we must encode the binary data to base64 and then decode it upon receive, base64 usage implies:
- more space on the wire (+33%)
- more processing needed

Hamdy:
fixes #63


specs/raml/sdstor.raml
```diff
@@ -114,12 +114,18 @@ types:
...
+        description: |
+          Key that is used to retrieve the object back again. Object ids are unique per namespace.
         required: true
         type: string
         minLength: 1
         maxLength: 128
         pattern: ^\w+$
       data:
+        description: |
+            Contents of the actual object uploaded. It's up to client to determine the type of object data.
```
zaibon: this is no good
we have to explain the two different format expected depending if we use multi/part or full json
now it looks like it doesn't matter how you format the data

server/api/rest/namespacesnsid_api_CreateObject.go
```diff
@@ -1,35 +1,107 @@
...
 	"github.com/zero-os/0-stor/server/stats"
+	"net/http"
+	"strings"
 )
 
+// parse multipart form
+// It expects a request with (Multi part form)
+// Form fields are : id (string), referenceList(string/json array), data(binary)
+func ParseMultipartForm(r *http.Request) (BinaryObject, error) {
+	emptyObj := BinaryObject{}
```
zaibon: Why do you create this object, and then you don't use it at all ?

you could use it when you read data from the form
e.g.: 
```go
emptyObj.ReferenceList = r.FormValue("referenceList")
```

Like that you allocate only one object.

server/api/rest/namespacesnsid_api_CreateObject.go
```diff
@@ -1,35 +1,107 @@
...
+		// populate reqBody with request data
+		reqBody.Id = body.Id
+		reqBody.Data = body.Data
+		for _, v := range strings.Split(body.ReferenceList, ","){
+			reqBody.ReferenceList = append(reqBody.ReferenceList, ReferenceID(v))
+		}
 	}
 
+	// increase request counter
+	go stats.IncrWrite(mux.Vars(r)["nsid"])
```
zaibon: this need to be at the beginning of the function, cause you can return before this line, that would mean you've not counted this request

server/api/rest/namespacesnsid_api_GetObject.go
```diff
@@ -34,22 +35,48 @@ func (api NamespacesAPI) GetObject(w http.ResponseWriter, r *http.Request) {
...
-		resp.ReferenceList = append(resp.ReferenceList, refid)
-	}
 
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
-	json.NewEncoder(w).Encode(resp)
+
+		json.NewEncoder(w).Encode(resp)
+	}else {
+		resp := BinaryObject{
```
zaibon: Shouldn't this return a multipart form and not a json object ?

zaibon:
regarding this issue https://github.com/zero-os/0-stor/issues/191, I'll close this one.

PR #148, opened 2017-08-11
Add test to client 

zaibon:


client/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// func testGRPCServer(t *testing.T, n int) (storserver.StoreServer, *itsyouonline.Client, string, func()) {
+func testGRPCServer(t testing.TB, n int) ([]storserver.StoreServer, func()) {
+	servers := make([]storserver.StoreServer, n)
+	dirs := make([]string, n)
+
+	for i := 0; i < n; i++ {
+
+		tmpDir, err := ioutil.TempDir("", fmt.Sprintf("0stortest-%d", i))
```
iwanbk: >  fmt.Sprintf("0stortest-%d", i)

I think we don't really need formatting here, it is going to be deleted quickly anyway

client/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+	defer func() {
+		etcdClean()
+		serverClean()
+	}()
+
+	shards := make([]string, len(servers))
+	for i, server := range servers {
+		shards[i] = server.Addr()
+	}
```
iwanbk:
```
shards := make([]string, 0, len(servers))
for _, server := range servers {
 		shards = append(shards, server.Addr())
}
```
I prefer this way, safer

client/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+			for i := 0; i < b.N; i++ {
+				key := []byte("testkey")
+				_, err := c.Write(key, data, nil, nil)
+				require.NoError(b, err, "fail to write data to the store")
+			}
+		})
+	}
+}
+
+func isIn(target string, list []string) bool {
```
iwanbk: i think we can simply use `assert.Contains` or `require.Contains`, not sure, godoc is down from my place at the moment

client/lib/distribution/stor_distribution.go
```diff
@@ -89,23 +89,31 @@ func (sd StorDistributor) WriteBlock(key, value []byte, md *meta.Meta) (*meta.Me
...
 			}
-		}(i, piece)
+		}(i, piece, cErr)
 	}
+
+	go func(cErr chan error) {
+		for err := range cErr {
+			errs = append(errs, err)
+		}
+	}(cErr)
```
iwanbk: nice trick

client/lib/encrypt/aes.go
```diff
@@ -37,15 +35,24 @@ func newAESGCM(privKey, nonce []byte) (*aesgcm, error) {
...
+	return ag.gcm.Seal(nonce, nonce, plain, nil), nil
 }
 
 func (ag *aesgcm) Decrypt(cipher []byte) ([]byte, error) {
-	return ag.gcm.Open(nil, ag.nonce, cipher, nil)
+	if len(cipher) < ag.gcm.NonceSize() {
+		return nil, errors.New("malformed ciphertext")
+	}
+
+	return ag.gcm.Open(nil, cipher[:ag.gcm.NonceSize()], cipher[ag.gcm.NonceSize():], nil)
```
iwanbk: :+1:

client/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/coreos/etcd/embed"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/zero-os/0-stor/client/config"
+	"github.com/zero-os/0-stor/client/itsyouonline"
+	"github.com/zero-os/0-stor/client/lib/compress"
+	"github.com/zero-os/0-stor/server/storserver"
+)
+
+func testStartEtcd(t testing.TB) (string, func()) {
```
iwanbk: I think we need to remove this func

client/client_test.go
zaibon: but we can't start all the server using the same data directory, that would open the same badger file multiple time which for sure will cause corruption

client/client_test.go
zaibon: but slower, and I don't see why it's not safe here, I allocate the right size and range over the slice so no risk of out of range

client/client_test.go
iwanbk: `ioutil.TempDir` already guarantee for unique result

https://golang.org/pkg/io/ioutil/#TempDir
```
Multiple programs calling TempDir simultaneously will not choose the same directory
```

client/client_test.go
iwanbk: yep, it is safe here

client/client_test.go
zaibon: good point, didn't know that

PR #167, opened 2017-08-18
Add meta client timeout on Dial, Put, and Get

### LINKED ISSUE #165: cli is hanging in case of etcd is off
**Steps**
- Turn off etcd
- Try to upload file

**Actual result**
image

iwanbk:
Fixes #165


client/meta/client.go
```diff
@@ -37,13 +44,17 @@ func (c *Client) Put(key string, meta *Meta) error {
 	if err := meta.Encode(buf); err != nil {
 		return err
 	}
-	_, err := c.etcdClient.Put(context.TODO(), key, string(buf.Bytes()))
+	ctx, _ := context.WithTimeout(context.Background(), metaOpTimeout)
```
zaibon: I don't think you can omit the cancel function. Context package requires to always call the cancel function to release resources of the context. Not doing it will create goroutine leaks

client/meta/client.go
```diff
@@ -37,13 +44,17 @@ func (c *Client) Put(key string, meta *Meta) error {
...
+
+	_, err := c.etcdClient.Put(ctx, key, string(buf.Bytes()))
 	return err
 }
 
 // Get fetch metadata from metadata server
 func (c *Client) Get(key string) (*Meta, error) {
-	resp, err := c.etcdClient.Get(context.TODO(), key)
+	ctx, _ := context.WithTimeout(context.Background(), metaOpTimeout)
```
zaibon: same comment

client/meta/client.go
iwanbk: thanks, didn't know that, fixed

client/meta/client.go
iwanbk: fixed

Issue #169, opened 2017-08-18
Rename main packages to something more meaningful 


zaibon:
Currently the  package name and binaries produces by this repository are too generic (server, client, cli)

For the binary I propose we had a makefile that build into a bin directory with proper name. `zerostorserver` and `zerostorcli` to be consistent with the naming of the 0disk project.

And for the client package, we should also rename it to zerostorclient or something. Don't really like this name either but I have no better idea for now

GlenDC:
The name should still be defined by the directory as that's go idiomatic (please do not define the name using `-o <name>` during builds, as that would still make `go get` installs fail. Often you have a root `/cmd` package, with there all your binaries in their separate directories. Shared code would be in other root packages (or sub packages of those root packages), while code-specific code would be in the binaries (sub)package.

e.g. https://github.com/kubernetes/kubernetes/tree/master/cmd

So for 0-stor that would mean something like

+ `github.com/zero-os/0-stor/cmd/zerostorctl`
+ `github.com/zero-os/0-stor/cmd/zerostorserver`

Calling the client package client is fine, you could also call it api, but either way both names or generic. This is really ok. Your name is already made unique by the packages in front of it (e.g. `github.com/zero-os/0-stor/client` or `github.com/zero-os/0-stor/api`).

It's rare that when using such client/api files that you have package conflicts, and for those case where you do Golang allows you to define a custom name for it as part of your import, so that in such cases people can still name it `zerostoclient` or `zerostorapi` (while importing) if they want.

zaibon:
Yep make sense.
We already have a cmd package in the client. I'll just move it at the root and create another package for the server binary.

PR #171, opened 2017-08-18
make building of binaries more easy

zaibon:
Part of #169


Makefile
```diff
@@ -0,0 +1,55 @@
...
+TIMEOUT ?= 10m
+
+PACKAGE = github.com/zero-os/0-stor
+COMMIT_HASH = $(shell git rev-parse --short HEAD 2>/dev/null)
+BUILD_DATE = $(shell date +%FT%T%z)
+
+SERVER_PACKAGES = $(shell go list ./server/...)
+CLIENT_PACKAGES = $(shell go list ./client/...)
+
+ldflags = -extldflags "-static" -s -w
```
iwanbk: what is the meaning of this? and why we need it

Makefile
```diff
@@ -0,0 +1,55 @@
...
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) \
+		go build -ldflags '$(ldflags)' -o $(OUTPUT)/$@ ./client/cmd/cli
+endif
+
+server: $(OUTPUT)
+ifeq ($(GOOS), darwin)
+	GOOS=$(GOOS) GOARCH=$(GOARCH) \
+		go build -o $(OUTPUT)/$@ ./server
+else
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) \
```
iwanbk: why we need `CGO_ENABLED` var?

docs/gettingstarted.md
```diff
@@ -1,11 +1,54 @@
...
 
-**Download**
+The server is go-getable :
 ```
 go get -u github.com/zero-os/0-stor/server
 ```
+This will create `$GOPATH/bin/server`
+
+
+Since server is not really explicty the recommanded method to build is to use the makefile at the root of the repository
```
iwanbk: > recommanded method to build is to use the makefile at the root of the repository

Need to describe more why `make` is the recommended one. I think `go get` is the easier & safer (because it always give us latest code) but slower (need to check the repo for latest code)

Makefile
zaibon: It make sure that we build fully static binary. We use static binary so the creation of the flist for this binary is easier since we can put only the binary in the flist and no dependent libraries

Makefile
zaibon: It's needed for the creation of fully static binary

docs/gettingstarted.md
zaibon: that's what I do here. go get will generate a binary called `server`, this is too generic name.

iwanbk: Oh my bad, perhaps fixing the typo will make it better

PR #184, opened 2017-08-23
Let user specify own initial metadata on Write

iwanbk:
For example to specify own epoch time.

Specifying own epoch time is needed by tlog, so the 0-stor client `Walk` is going to use same epoch time as tlog aggregation


client/client.go
```diff
@@ -73,10 +73,11 @@ func newClient(conf *config.Config) (*Client, error) {
 // WriteF writes the key with value taken from given io.Reader
 // it currently only support `chunker` as first pipe.
 // Metadata linked list will be build if prevKey is not nil.
-// prevMeta is previous metadata, to be used in case of user already has the prev meta.
+// prevMeta is optional previous metadata, to be used in case of user already has the prev meta.
 // So the client won't need to fetch it back from the metadata server.
 // prevKey still need to be set it prevMeta is set
-func (c *Client) WriteF(key []byte, r io.Reader, prevKey []byte, prevMeta *meta.Meta) (*meta.Meta, error) {
+// curMeta is optional metadata, if user want to set his own initial metadata for example set own epoch
+func (c *Client) WriteF(key []byte, r io.Reader, prevKey []byte, prevMeta, curMeta *meta.Meta) (*meta.Meta, error) {
```
zaibon: Is curMeta a good name if it's only use as initial Metadata?

iwanbk: you're right, changed to `initialMeta`

PR #186, opened 2017-08-23
Create client specification

iwanbk:
I know it is too late, but better than no spec


zaibon:
Better to make this a documentation instead of specs since the features are already there.
Normal flow should be `specs` -> `implementation` -> `specs are converted to docs`

iwanbk:
agree

PR #189, opened 2017-08-23
Distribution return failed servers in case of error.

### LINKED ISSUE #187: distribution should return failed servers
as described in principles

iwanbk:
Fixes #187


client/lib/distribution/stor_distribution.go
```diff
@@ -101,7 +101,7 @@ func (sd StorDistributor) WriteBlock(key, value []byte, md *meta.Meta) (*meta.Me
 			defer wg.Done()
 			if _, err := sd.storClients[idx].ObjectCreate(hashedKey, data, nil); err != nil {
 				mux.Lock()
-				errs = append(errs, err)
+				errs = append(errs, fmt.Errorf("%v", sd.shards[idx]))
```
zaibon: shouldn't we put both error message and failed servers ?

iwanbk: was thinking the same, and i think it is better

iwanbk: I think the error message should be easy to parse because it means to be consumed by:
- the user of this library, in case of Go client
- user of client daemon, python or other languages

Both kind need to log the error and report the failed server to other part (e.g. tlog reports to 0-orch).

JSON is good enough format i think, easy to parse.
It is true that it is much slower than capnp, but error handling is not in hot path

iwanbk:
```
I think the error message should be easy to parse because it means to be consumed by:

the user of this library, in case of Go client
user of client daemon, python or other languages
Both kind need to log the error and report the failed server to other part (e.g. tlog reports to 0-orch).

JSON is good enough format i think, easy to parse.
It is true that it is much slower than capnp, but error handling is not in hot path
```
(Moved from https://github.com/zero-os/0-stor/pull/189#discussion_r134719058, because github code comment might get collapsed on code update)

The potential problem with JSON is that JSON marshall (needed by `Error` func implementation) could return an error, which will be pretty silly if it really happen in func that supposed to print the error value it contains
There are two kind of errors that might happen:
- [UnsupportedTypeError](https://golang.org/pkg/encoding/json/#UnsupportedTypeError) 
- [UnsupportedValueError](https://golang.org/pkg/encoding/json/#UnsupportedValueError)

Because we use struct with non interface fields, i think we could prevent the error to happens.
And it is pretty safe to use JSON here.

What do you think @zaibon ?

The error will contain:
- shard address
- type of shard (etcd, 0-stor, ...)
- error message
- error code (if we could classify the error)

zaibon:
Proposition looks good. You can proceed with it

iwanbk:
> This error struct (might be called ShardError) could also be used by replication

The `ShardError` has been implemented at e12c4cc57d3f27414214fc9ebb20ef4a8f465a31

I'm continuing the work for replication

iwanbk:
@zaibon the PR is ready now  :)

PR #190, opened 2017-08-23
Pipe component should never creates own token.

### LINKED ISSUE #188: There are multiple token creations in client
Should be reduced to one

iwanbk:
The client will be supplied by main client.

Fixes #188


client/client.go
```diff
@@ -38,22 +39,33 @@ func New(conf *config.Config) (*Client, error) {
 }
 
 func newClient(conf *config.Config) (*Client, error) {
+	nsMgr := itsyouonline.NewClient(conf.Organization, conf.IYOAppID, conf.IYOSecret)
+	iyoToken, err := nsMgr.CreateJWT(conf.Namespace, itsyouonline.Permission{
+		Write: true,
```
zaibon: We should ask for all permission, IYO will only return the scope that the user actually has anyway.

iwanbk: Why all? we only have read & write pipe

zaibon:
https://github.com/zero-os/0-stor/issues/188 make sense, but what about the discussion regarding allowing unit test without the need of a jwt. With this PR, I don't think we make a move in that direction.

iwanbk:
> make sense, but what about the discussion regarding allowing unit test without the need of a jwt. With this PR, I don't think we make a move in that direction.

This PR also wont harm the path to that direction

iwanbk:
> This PR also wont harm the path to that direction

We actually move to that direction with this path, because the client only creates the token once. It means we only have one place to modify in case of we allow empty token

zaibon:
Make sense.

Please rebase from master, seems the test fails since the change on the travis file on master

iwanbk:
> Please rebase from master, seems the test fails since the change on the travis file on master

done

client/client.go
zaibon: Inded, and that make me thing that we miss some method on the client then. 
Cause we also should be able to delete objects. But let's do that in another PR

Issue #180, opened 2017-08-22
Update CLI to return the metadata after uploading a file


zaibon:
Like the go client API has, the CLI should also return the metaadata created after a write operation.

iwanbk:
What for? to print it to the console?

zaibon:
Yep. The problem is currently QA is writing test scenarios. They use python for their work. Since we don't have python client yet,they only use the cli. But the cli doesn't return any Metadata. So it's hard for them to implement proper tests since they don't know where the data has been put. 

So or we return the Metadata after each upload. Or we add a command that return the Metadata for a certain key. The output should be something easily parsable to make their life easier

zaibon:
Yep. The problem is currently QA is writing test scenarios. They use python for their work. Since we don't have python client yet,they only use the cli. But the cli doesn't return any Metadata. So it's hard for them to implement proper tests since they don't know where the data has been put. 

So or we return the Metadata after each upload. Or we add a command that return the Metadata for a certain key. The output should be something easily parsable to make their life easier

islamTaha12:
closed
```bash
➜  zerostorcli git:(master) ✗ zerostorcli file metadata config.yaml | json_pp
{
   "previous" : null,
   "key" : "Y29uZmlnLnlhbWw=",
   "epoch" : 1504783444508130741,
   "next" : null,
   "encryption_key" : "YWIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTI=",
   "chunks" : [
      {
         "key" : "DSEAjJd6uLyX4gnVTxm2vGLLSoWopjZvEo//IzPCWC4=",
         "shards" : [
            "http://10.147.17.205:8080",
            "http://10.147.17.66:8080",
            "http://10.147.17.234:8080",
            "http://10.147.17.85:8080"
         ],
         "size" : 783
      }
   ],
   "configPtr" : null
}

```

Issue #135, opened 2017-08-09
Reduce the need to write to metadata server


iwanbk:
We currently write to metadata server for each block we upload to 0-stor server.
It is something unavoidable (at least for now) if that block is not part of bigger entity (i.e. the block is part of bigger block).

We might be able to write only one metadata for multiple blocks that is part of bigger entity (e.g. big file chunked into multiple chunks).
However, it will conflict with #126

iwanbk:
The first step we can do (~if we really need to do it~) is to make the meta pipeline explicit again or at least the config module added it automatically if it doesn't see any components which write to metadata.

So, the distribution & stor.client component doesn't write to the metadata server.

zaibon:
I have another propostion, cause I don't see how the `meta pipeline` would solve anything.

We need to change the capnp schema of the metadata. 
Currently it looks like this
```capnp
struct Metadata {
	size @0 :UInt64;
	# Size of the data in bytes

	epoch @1 :UInt64;
	# creation epoch

	key @2 :Data;
	# Key used in 0-stor

	encrKey @3 :Data;
	# Encryption key used to encrypt this file

	shard @4 :List(Text);
	# List of shard of the file. It's a url the 0-stor

	previous @5 :Data;
	# Key to the previous metadata entry

	next @6 :Data;
	# Key to the next metadata entry

	configPtr @7 :Data;
	# Key to the configuration used by the lib to set the data.

	numOfChunks @8:UInt64;
	# number of chunks
}
```

I propose to change it to 
```capnp
struct Metadata {
	size @0 :UInt64;
	# Size of the data in bytes

	epoch @1 :UInt64;
	# creation epoch

	key @2 :Data;
	# Key used in 0-stor

	encrKey @3 :Data;
	# Encryption key used to encrypt this file

	struct Chunk {
		key @0 :Data;
		# Key used in 0-stor

		size @1 :UInt64;
		# Size of the chunk in bytes

		Shards @2 :List(Text);
		# List of shards of the chunk. It's a url the 0-stor
	}
	chunks @4 :List(Chunk);
	# List of chunks of the file.
	# Can be only 1 if we don't split the file with chunker pipe

	previous @5 :Data;
	# Key to the previous metadata entry

	next @6 :Data;
	# Key to the next metadata entry

	configPtr @7 :Data;
	# Key to the configuration used by the lib to set the data.
}
```

[...]

PR #217, opened 2017-08-30
Refactoring metadata + allow to get metadata of a file form cli

zaibon:
Relates to :
- https://github.com/zero-os/0-stor/issues/180
- https://github.com/zero-os/0-stor/issues/135


client/client.go
```diff
@@ -177,18 +179,14 @@ func (c *Client) Read(key []byte) ([]byte, error) {
 
 func (c *Client) linkMeta(curMd, prevMd *meta.Meta, curKey, prevKey []byte) error {
 	if len(prevKey) == 0 {
-		return nil
+		return c.metaCli.Put(string(curKey), curMd)
```
iwanbk: Why we need this line now?

client/cmd/zerostorcli/main.go
```diff
@@ -58,11 +68,15 @@ func main() {
 							return cli.NewExitError(fmt.Errorf("can't read the file: %v", err), 1)
 						}
 
-						if _, err := cl.WriteF([]byte(fileName), f, nil, nil, nil); err != nil {
-							return cli.NewExitError(fmt.Errorf("upload failed : %v", err), 1)
+						if key == "" {
+							key = path.Base(fileName)
```
iwanbk: why not using https://golang.org/pkg/path/filepath/ instead of `path`?

client/lib/distribution/stor_distribution.go
```diff
@@ -64,13 +64,15 @@ func (sd StorDistributor) WriteBlock(key, value []byte, md *meta.Meta) (*meta.Me
 	}
 
 	var (
-		shardErr lib.ShardError
+		shardErr = &lib.ShardError{}
```
iwanbk: why need to change it?

client/lib/distribution/stor_distribution.go
```diff
@@ -216,7 +201,7 @@ func (sr StorRestorer) ReadBlock(metaKey []byte) ([]byte, error) {
 	}
 
 	// decode
-	return sr.dec.Decode(chunks, int(md.Size()))
+	return sr.dec.Decode(shards, int(md.Size()))
```
iwanbk: I think `shards` is not a good name here, because we usually use it for server shards.

It tricked me actually :)

`pieces` is better i think

client/README.md
```diff
@@ -38,20 +38,34 @@ API documentation : [https://godoc.org/github.com/zero-os/0-stor/client](https:/
...
-    }
-
+	type Chunk struct {
+		Size   uint64    # Size of the chunk in bytes
+		Key    []byte    # key used in the 0-stor
+		Shards []string
+	}
+	type Meta struct {
+		Epoch     int64  # creation epoch
+		Key       []byte # key used in the 0-stor
```
iwanbk: i think it could be removed

client/lib/block/bytesbuff.go
```diff
@@ -20,7 +20,6 @@ func NewBytesBuffer() *BytesBuffer {
 
 // WriteBlock implements block.Writer interface
 func (bb *BytesBuffer) WriteBlock(key, val []byte, md *meta.Meta) (*meta.Meta, error) {
-	n, err := bb.Buffer.Write(val)
-	md.SetSize(uint64(n))
```
iwanbk: why deleted?

client/lib/compress/compress_test.go
```diff
@@ -42,23 +40,22 @@ func testRoundTrip(t *testing.T, conf Config) {
...
+	md := meta.New(nil)
 	// compress by write to the writer
 	_, err = w.WriteBlock(nil, payload, md)
-	assert.Nil(t, err)
+	assert.NoError(t, err)
 
 	// create reader
 	r, err := NewReader(conf)
-	if !assert.Nil(t, err) {
+	if !assert.NoError(t, err) {
```
iwanbk: Curious, what if the difference between NoError and Nil?

client/lib/encrypt/encrypt.go
```diff
@@ -42,8 +42,9 @@ func NewEncrypterDecrypter(conf Config) (EncrypterDecrypter, error) {
 
 // Writer defines encryption writer
 type Writer struct {
-	ed EncrypterDecrypter
-	w  block.Writer
+	ed      EncrypterDecrypter
+	w       block.Writer
+	encrKey string
```
iwanbk: i wonder why we don't need to store `encrKey` before. Was it a bug?

client/lib/replication/writer.go
```diff
@@ -59,7 +58,7 @@ func NewStorWriter(w block.Writer, conf Config, shards, metaShards []string, org
 // If number of failed shards is more than max failed allowed,
 // then the error contains the failed shards
 func (sw *StorWriter) WriteBlock(key, val []byte, md *meta.Meta) (*meta.Meta, error) {
-	var shardErr lib.ShardError
+	var shardErr *lib.ShardError
```
iwanbk: why need to change to pointer?

client/client.go
zaibon: This method is the only place where we write Metadata to the etcd. So we always need to write even if we don't use linked list

client/lib/distribution/stor_distribution.go
zaibon: Cause the object contains a mutex and you can't pass mutex per copy. It gives a go vet error if you don't use a pointer

client/lib/distribution/stor_distribution.go
zaibon: Agree

client/lib/encrypt/encrypt.go
zaibon: I need to be able to get key in the because I set the encryption key to the Metadata during WriteBlock

client/cmd/zerostorcli/main.go
zaibon: indeed

client/README.md
zaibon: This is still used in some pipe to get the metadata object

client/lib/block/bytesbuff.go
zaibon: There is no size on the metadata object anymore. Instead I have a Size() method that does the sum of all the chunks sizes

client/lib/compress/compress_test.go
zaibon: assert lib gives better explanation for error when you test for error using `NoError` then `Nil`

client/client.go
iwanbk: No, the the pipe which write to 0-stor should also write it, at least in old code

client/lib/distribution/stor_distribution.go
iwanbk: oh my bad

PR #221, opened 2017-09-06
Run test suite

islamTaha12:


zaibon:
included in https://github.com/zero-os/0-stor/pull/223

PR #223, opened 2017-09-07
Refactor client

zaibon:
Big refactoring of the client.

- Remove the concept of pipes. instead simpler client policy define how the data processing occurs.
- Also makes client public API much more cleaner. 
- Supports for streaming of data

Also include the QA test suite


client/client.go
```diff
@@ -87,96 +101,543 @@ func (c *Client) Close() error {
...
+			md.EncrKey = []byte(c.policy.EncryptKey)
+			chunk.Size = uint64(len(block))
+		}
+
+		if c.policy.Compress {
+			block = snappy.Encode(nil, block)
+			chunk.Size = uint64(len(block))
+		}
+
+		if c.policy.ReplicationMaxSize > 0 && len(block) <= c.policy.ReplicationMaxSize {
```
iwanbk: better to move this logic to method of `policy`, so it is more clear about what it does and can be reused, 

I saw it being used in somewhere else

client/client.go
```diff
@@ -87,96 +101,543 @@ func (c *Client) Close() error {
...
+			usedShards, err = c.replicateWrite(chunkKey, block, []string{})
+			if err != nil {
+				return nil, err
+			}
+			chunk.Size = uint64(len(block))
+
+		} else if c.policy.DistributionNr > 0 && c.policy.DistributionRedundancy > 0 {
```
iwanbk: better to move this logic to method of policy, so it is more clear about what it does and can be reused,

I saw it being used in somewhere else

client/client.go
```diff
@@ -87,96 +101,543 @@ func (c *Client) Close() error {
...
+			usedShards, _, err = c.distributeWrite(chunkKey, block, []string{})
+			if err != nil {
+				return nil, err
+			}
+			chunk.Size = uint64(len(block))
+		} else {
```
iwanbk: perhaps `switch` is better than `if else` here

client/client.go
```diff
@@ -87,96 +101,543 @@ func (c *Client) Close() error {
...
+			}
+
+		} else if c.policy.DistributionNr > 0 && c.policy.DistributionRedundancy > 0 {
+			block, err = c.distributeRead(chunk.Key, int(chunk.Size), chunk.Shards)
+			if err != nil {
+				return err
+			}
+		} else {
+			if len(chunk.Shards) <= 0 {
+				panic("metadata corrupted, can't have a chunk without shard")
```
iwanbk: why `panic` instead of returning error?

I think it won't play well with our upcoming daemon

PR #234, opened 2017-09-12
Add referenceList as Write* argument.

iwanbk:
Without this, client need two API calls to create object with reference list


client/client_test.go
```diff
@@ -229,7 +229,7 @@ func TestRoundTripGRPC(t *testing.T) {
 
 			// write data to the store
 			key := []byte("testkey")
-			meta, err := c.Write(key, data)
+			meta, err := c.Write(key, data, nil)
```
zaibon: Instead of passing nil, pass a list and check its the same after read.

cmd/zerostorcli/main.go
```diff
@@ -99,7 +99,7 @@ func main() {
 							key = filepath.Base(fileName)
 						}
 
-						_, err = cl.WriteF([]byte(key), f)
+						_, err = cl.WriteF([]byte(key), f, nil)
```
zaibon: Here we should add a flag that allow to pass reference list instead if just always passing nil

client/client_test.go
iwanbk: I was thinking the same, but `Read` API doesn't return the `reference list`.

I think it is better to go with variadic argument, right?

zaibon: I see 2 options.

Or we also return the reference list in Read methods, or instead we return an object that hold, key,value and reference list.

For the Write variadic is an option but then user have to pass multiple string instead of the slice of string, which I don't like that much. So I think we can still break the API at this stage.  Since we are the only consumer so far.

iwanbk: > or instead we return an object that hold, key,value and reference list.

I prefer this way.

Returning three variables is a bit too much, not sure.
Four is certainly too much, not sure about three.

> For the Write variadic is an option but then user have to pass multiple string instead of the slice of string, which I don't like that much.

agree

zaibon: > I prefer this way.

Me too, let's go for this

iwanbk:
@zaibon i implemented your remarks, except returns value of `Read*`.
Instead of returning object as we agreed, i added `referenceList` to the returned value.
It is because doesn't make sense for `ReadF` to returns an object.

Another note is reference list when chunker is used.
In theory, each chunk could have difference reference list, not sure in reality.
In this PR, i simply use reference list of last chunk

client/client.go
```diff
@@ -432,9 +433,9 @@ func (c *Client) replicateWrite(key, value []byte, referenceList []string) ([]st
 	return usedShards, nil
 }
 
-func (c *Client) replicateRead(key []byte, shards []string) ([]byte, error) {
+func (c *Client) replicateRead(key []byte, shards []string) (*pb.Object, error) {
```
zaibon: Shouldn't you return ([]byte,[]string,error) here to be consistent with the rest?

client/client.go
```diff
@@ -593,7 +597,12 @@ func (c *Client) distributeRead(key []byte, originalSize int, shards []string) (
 				shardErr.Add([]string{shard}, lib.ShardType0Stor, err, 0)
 				return
 			}
+			mux.Lock()
```
zaibon: Let's try to find a way to avoid this lock. I made the method as such that we don't need lock in the first place, I would like to keep it like that.

client/client_test.go
```diff
@@ -245,13 +245,13 @@ func TestRoundTripGRPC(t *testing.T) {
...
 			// read data back
-			dataRead, err := c.Read(key)
+			dataRead, refListRead, err := c.Read(key)
 			require.NoError(t, err, "fail to read data from the store")
 			if bytes.Compare(data, dataRead) != 0 {
 				t.Errorf("data read from store is not the same as original data")
 				t.Errorf("len original: %d len actual %d", len(data), len(dataRead))
 			}
-			// assert.Equal(t, data, dataRead)
+			require.Equal(t, refList, refListRead)
```
zaibon: you can also test the data `assert.Equal(t, data, dataRead)`

client/client.go
iwanbk: the other one is `distributeRead` which returns object.
but make it consistent with the API might be better

client/client.go
iwanbk: actually i wonder the test race didn't catch the old code.
it could access the `slice` in the same time.
perhaps because it always write to different index.
But with `map`, write to different key in the same time will raise an error

client/client.go
zaibon: ha yes `distributeRead` also returns object now.

client/client_test.go
iwanbk: i keep the old code because in case of error, `assert.Equal(t, data, dataRead)` will pollute our screen with garbage. Two 4096 bytes binary string will be displayed on our screen

client/client.go
zaibon: The race was not possible cause you never write to the same index at the same time, that was there all the trick to avoid a mutex.

There is not map used here, so this is valid

client/client_test.go
zaibon: I think assert lib don't print the content when the slice is too big.

client/client.go
iwanbk: > There is not map used here, so this is valid

I know, it was just for comparison.

from quick search, it seems you're right.
http://grokbase.com/t/gg/golang-nuts/14138jw64s/go-nuts-concurrent-read-write-of-different-parts-of-a-slice#201401054rz5g2775ebxfldaqdavgdoyce

I can  see two options here:
- create slice of reflist, then each goroutine write to respective index
- create channel of refList with size == number of gorutines.
I don't really like it because it is a waste of memory to store many values but only use on of it.

What do you think?

client/client_test.go
iwanbk: it does, i just tried it.

zaibon: ok no need to add it then

client/client.go
iwanbk: fixed, i go with 1st option

iwanbk:
@zaibon i addressed your remark about usage of closure.

PR #237, opened 2017-09-13
Allow to start the 0-stor server with JWT authentification disabled

zaibon:
This relates to : https://github.com/zero-os/0-Disk/issues/457


server/db/badger/badger.go
```diff
@@ -127,9 +128,10 @@ func (b BadgerDB) Filter(prefix []byte, start int, count int) ([][]byte, error)
 		item := it.Item()
 
 		var val []byte
-		err := item.Value(func(v []byte) {
+		err := item.Value(func(v []byte) error {
 			val = make([]byte, len(v))
 			copy(val, v)
+			return nil
```
iwanbk: It seems this func never returns error?

zaibon: it doesn't but there is not error that can be raised here. copy only return an int.

Issue #240, opened 2017-09-15
goroutine leak in client replicateRead


iwanbk:
`replicateRead` https://github.com/zero-os/0-stor/blob/69565d52f2b4ea4673a9484d1383350144a0fe6f/client/client.go#L441-L492 has goroutine leaks bug.

it is because of this part of code
```go
			select {
			case <-cQuit:
				return
			default:
				obj, err := cl.ObjectGet(key)
				if err != nil {
					log.Warningf("replication read, error reading from %s: %v", shard, err)
					return
				}
				cVal <- obj
			}
```
The `select` supposed to return in case of `cQuit` being closed, but this won't happen in case of it already in `cl.ObjectGet`

Check it [in playground](https://play.golang.org/p/ttI6T25T__) to see it. I simulate `cl.ObjectGet` with `time.Sleep`
I don't send PR to fix it because it seems you don't really want for all servers being hit by the client.
The solutions i can see is either doing it sequentially or concurrently to all servers.

I prefer concurrently, for speed reason.

I could send a PR if you agree @zaibon

zaibon:
This is right. I wanted to avoid we do a call to the store if another one already answered, but my code is buggy indeed.
We should do the `ObjectGet` call before the select.

Issue #225, opened 2017-09-07
Text file with repeated content failed to be downloaded


islamTaha12:
```bash
➜  zerostorcli git:(master) ✗ zerostorcli file upload /home/upload/upload_file_106                    
file uploaded, key = upload_file_106
➜  zerostorcli git:(master) ✗ zerostorcli file download upload_file_106 /home/upload/upload_file_106_1
download file failed: snappy: corrupt input
➜  zerostorcli git:(master) ✗ 
```

**Config:**
```bash
organization: xtremx90 # itsyou.online organization name
namespace: namespace_test       # itsyou.online namespace
protocol: rest
iyo_app_id:
iyo_app_secret: 
# the address(es) of 0-stor server
data_shards:
    - http://10.147.17.234:8080
    - http://10.147.17.66:8080
    - http://10.147.17.205:8080
    - http://10.147.17.85:8080
# the address(es) of etcd server
meta_shards:
    - http://10.147.17.234:2379

block_size: 4096

replication_nr: 4
replication_max_size: 4096

distribution_data: 3
distribution_parity: 1

compress: true
encrypt: true
encrypt_key: ab345678901234567890123456789012
```

**File size**
-rw-r--r-- 1 root root 89817 Aug 30 15:44 /home/upload/upload_file_106

**Affected version**
https://github.com/zero-os/0-stor/commit/a19fec7a8ec562c59134fd869e483f523f118844

zaibon:
Turns out the problem comes from the file that @islamTaha12 used to do his test.

For some reason the way he generated the file creates files that can't be downloaded.

zaibon:
Here is the file that is impossible to download
[test_file.zip](https://github.com/zero-os/0-stor/files/1285179/test_file.zip)

iwanbk:
i suspect that the chunker doesn't do the streaming properly

zaibon:
> i suspect that the chunker doesn't do the streaming properly

What makes you think that ?
Cause I tried to upload a file generated from /dev/uradom of the exact same size and it worked.

iwanbk:
I got similar issue, not sure if it is the same thing.

When i wrote unit test for new zeroctl copy, which copy tlog data from one vdisk to another, the code do the `Walk` twice.
First walk is OK, but the second walk got same error as this one
```
snappy: corrupt input 
```

zaibon:
Hum this is more serious then I thought then.
@iwanbk do you have the dataset that also make walk fails ?

[...]

PR #241, opened 2017-09-15
Implement Repair method on client

zaibon:


client/check.go
```diff
@@ -0,0 +1,103 @@
...
+			case <-ctx.Done():
+				// in case we already know something is corrupted, we don't need
+				// to check other blokcs
+				return
+
+			default:
+				checks, err := store.ObjectsCheck(ids)
+				if err != nil {
+					log.Errorf("error getting check status on shard %s: %v", shard, err)
+					cErr <- err
```
iwanbk: When checking for #240, i was sure that there is another code that has same bug.
Now i remember, this code.

In this case, if there are more than one error, only one goroutine could exit.
Subsequent goroutine will be blocked by the channel

client/check.go
```diff
@@ -0,0 +1,103 @@
...
+
+				}
+
+				for _, check := range checks {
+					if check.Status != CheckStatusOk {
+						// signal we found a corrupted or missing block
+						cStatus <- check.Status
```
iwanbk: same as bug with cErr above

client/check_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	meta, err := c.Write(key, data, []string{})
+	require.NoError(t, err, "fail write data")
+
+	// Check status is ok after a write
+	status, err := c.Check(meta.Key)
+	require.NoError(t, err, "fail to check object")
+	assert.Equal(t, CheckStatusOk, status)
+
+	// corrupt file by removing a block
```
iwanbk: need test for removing more than one block, make sure all goroutines could exit

client/client_test.go
```diff
@@ -429,6 +429,57 @@ func BenchmarkWriteFilesSizes(b *testing.B) {
 	}
 }
 
+func TestIssue225(t *testing.T) {
```
iwanbk: Is it mean #225 has been fixed? What is the actual problem?

client/repair.go
```diff
@@ -0,0 +1,183 @@
...
+	ErrAllReplicateCorrupted = fmt.Errorf("All the replicate are corrupted, repair impossible")
+	// ErrRepairSupport is returned when a block is not using replication or distribution
+	ErrRepairSupport = fmt.Errorf("block is not using replication or distribution, repair impossible")
+)
+
+// Repair repairs a broken file.
+// If the file is distributed and the ammout of corrupted chunks is acceptable,
+// we recreate the missing chunks.
+// Id the file is replicated and we still have one valid replicate, we create the missing replicate
+// till we reach the replication number configured in the policy
```
iwanbk: > till we reach the replication number configured in the policy

why not fixes all shards?

client/client_test.go
zaibon: That was a test to try to reproduce the #255 when we were still thinking that downloading multiple time the same file corrupted files. This test prove it doesn't.

client/repair.go
zaibon: you want to replicate randomly again, not using the same shards.

PR #244, opened 2017-09-18
fix policy validation for distribution and replication number

zaibon:


client/policy.go
```diff
@@ -108,12 +108,12 @@ func (p Policy) validate() error {
 		return ErrZeroMetaShards
 	}
 
-	if p.ReplicationNr > 0 && p.ReplicationNr < len(p.DataShards) {
+	if p.ReplicationNr > 0 && p.ReplicationNr > len(p.DataShards) {
```
iwanbk: I think we don't need this 
```go
p.ReplicationNr > 0
```
because `len(p.DataShards)` always `>0`

client/policy.go
```diff
@@ -108,12 +108,12 @@ func (p Policy) validate() error {
...
 		return ErrNotEnoughReplicationShards
 	}
 
 	distributionNr := (p.DistributionNr + p.DistributionRedundancy)
-	if distributionNr > 0 && distributionNr < len(p.DataShards) {
+	if distributionNr > 0 && distributionNr > len(p.DataShards) {
```
iwanbk: I think we don't need this 
```go
distributionNr > 0
```
because `len(p.DataShards)` always `>0`

Issue #250, opened 2017-09-22
race condition in badger


iwanbk:
Got race condition during `test -race`
- while working on #248 https://gist.github.com/iwanbk/5ce035d7d8e124d2c2c78d8479526bcb#file-gistfile1-txt

- in 0-disk travis test https://travis-ci.org/zero-os/0-Disk/jobs/278479303
   https://gist.github.com/iwanbk/5ce035d7d8e124d2c2c78d8479526bcb#file-in_0_disk_test-log

iwanbk:
> while working on #248 https://gist.github.com/iwanbk/5ce035d7d8e124d2c2c78d8479526bcb#file-gistfile1-txt

While working on it, i saw nothing suspicious in 0-stor server side.
Could you check it @zaibon ?

iwanbk:
> in 0-disk travis test https://travis-ci.org/zero-os/0-Disk/jobs/278479303
> https://gist.github.com/iwanbk/5ce035d7d8e124d2c2c78d8479526bcb#file-in_0_disk_test-log

It happens on this test
https://github.com/zero-os/0-Disk/blob/c1190fd2c2d9d132e831c320afcd1db2480de20d/tlog/copy/copier_test.go#L105-L106

which test this func
https://github.com/zero-os/0-Disk/blob/c1190fd2c2d9d132e831c320afcd1db2480de20d/tlog/copy/copier.go#L101-L140

iwanbk:
i tried to add sync.Mutex to zero-os/0-stor/server/db/badger/badger.go, not as fix attempt, only wanna see how it affect the issue.
We still have race condition :)

iwanbk:
adding `opts.SyncWrites = true` to the test server didn't fix the issue

zaibon:
I should you should report that on badger repo. I don't see anything fishy in our code either.

iwanbk:
An issue filled to badger repo with a way to reproduce it.

After some digging, i found that this issue started at this commit https://github.com/dgraph-io/badger/commit/46ceb47341cea03fd822fb81988baedacdd2f631

Moving back to older commit seems to fix this issue

PR #262, opened 2017-10-02
Update deps

zaibon:
The issue describe at #250 has been fixed in https://github.com/dgraph-io/badger/issues/243
So I updated badger lib to the last version again.


iwanbk:
travis build is failed

zaibon:
Yep can't see why though. Test in client packages reports failure but I don't know see any test that failed

iwanbk:
got same error in 0-disk https://travis-ci.org/zero-os/0-Disk/jobs/282651857

iwanbk:
i added some logging to the test, and it magically print the error

https://travis-ci.org/zero-os/0-Disk/jobs/285895697#L489
```
fatal error: runtime: out of memory
```

iwanbk:
profiling my test didn't produce enough result, as it only uses 99MB, mainly by `badger`.

After some digging, i found that the test won't fail if i skip the main part of the test, calling `DeleteWithMeta`.

It seems that this operation is heavy weight for badger, need to check it out more

zaibon:
@chrisvdg can you also check this branch and reduce the memory allocation where needed in our tests. 
Seems this is the same issue as #274

chrisvdg:
@zaibon , fixed, though definitely check if your tests are still doing what you want them to do!

iwanbk:
> @chrisvdg can you also check this branch and reduce the memory allocation where needed in our tests.
> Seems this is the same issue as #274 

Same comment to #274. the test passed for me when i reduce number of logs.
But i'm not sure if it is the right way to make the test passed, because i'm afraid that it means that the deletion will fail if number of data become big.
(in tlog case, it failed for 100 tlogs, passed for 25 or 50 logs).

On the other hand, travis machine is quite small ( 2GB of RAM). And this test never failed on my 16GB PC.

chrisvdg:
@iwanbk check the RAM consumption when testing with and without race detection, I doubt for production it will fail like it does in our testing as we won't run it with race detection ;)

iwanbk:
> , I doubt for production it will fail like it does in our testing as we won't run it with race detection ;)

Yeah, sure we won't.

I just wondering if the RAM usage will always increased with the number of logs increased?

iwanbk:
please forget my comment, i missed important remarks by @chrisvdg at https://github.com/zero-os/0-stor/issues/274#issuecomment-335831742

client/check_test.go
```diff
@@ -36,15 +36,15 @@ func TestCheck(t *testing.T) {
...
+		DistributionNr:         2,
 		DistributionRedundancy: 1,
 	}
 
 	c, err := getTestClient(policy)
 	require.NoError(t, err, "fail to create client")
 	defer c.Close()
 
-	data := make([]byte, 1204*10)
+	data := make([]byte, 602*10)
```
iwanbk: not really important comment, i just feel strange with `1204` and `602` here :D

PR #267, opened 2017-10-03
Raise the message size the grpc client can send/receive

zaibon:


iwanbk:
why we still need this? (i assume it related to previous PR). I believe previous PR already has unit test

zaibon:
I first though only the server need to configure the size of the messages,
but turns out the client also has to set it.

Problem is now in the 0-disk, we try to write bigger message, but the client doesn't support it.

PR #272, opened 2017-10-04
Changed created for the demo

zaibon:
- allow to configure grpc message maximum size
- new way to store reference list that allow faster write
- extract meta interface in prevision of replacement of etcd
- more tests


server/db/object.go
```diff
@@ -159,38 +203,116 @@ func (o *Object) RemoveReferenceList(refList []string) error {
...
-	return len(o.ReferenceList)
+	return len(refList), nil
+}
+
+func isNilReference(ref []byte) bool {
+	return bytes.Compare(ref, nilRefList[:]) == 0
+}
+
+// Validcrc compare the content of the data and the crc, return true if crc match data, false otherwrise
+func (o *Object) Validcrc() (bool, error) {
```
iwanbk: Is it also the old behaviour that we don't include refList when counting the crc?

zaibon: Yes it already was.
see https://github.com/zero-os/0-stor/pull/272/files/a4e22d8118a8367e11ab0267492cc50cc89a4781#diff-9ada5da89a3c8a3d625bec584ce56d64L195

Now that make me think maybe we could also have a CRC for the reference list now that it's store in it's own space.
And have the repair method also repair a reference list

PR #285, opened 2017-11-06
Replaces iyo user flag with email

### LINKED ISSUE #283: Update IYO login to use email address
Authenticating to IYO is now meant to be done by the users verified email address or phone number.
In 0-stor now we ask the user to provide it's username, but that is now hidden on IYO which can make it annoying for users and IYO client functions like `iyoClient.Organizations.GetInvitations` returns email addresses in the User field of the returned invites.

What needs to be done:
* rename `--user` flag to `--email`, we can already provide an email address in the `--user` flag atm so changing the flag name should be enough
* rename variables names
* update the docs to say that we need the IYO email address

chrisvdg:
Closes #283


cmd/zerostorcli/README.md
```diff
@@ -135,10 +135,12 @@ This command will delete the organization `myorg.0stor.namespace_test` and all i
...
 
 ```
-./zerostorcli --conf conf_file.yaml namespace set-acl --namespace namespace_test --user johndoe -r -w -d
+./zerostorcli --conf conf_file.yaml namespace set-acl --namespace namespace_test --userid johndoe@email.com -r -w -d
 ```
 
-This command will authorize the user `johndoe` into the namespace `namespace_test` with the right read, write and delete.
+This command will authorize the user with it's [ItsYou.online][iyo] user ID, in this case the email address `johndoe@email.com` into the namespace `namespace_test` with the right read, write and delete.
+
+For [ItsYou.online's][iyo] user ID following can be used reliably with 0-stor: email address
```
zaibon: I think this sentence requires a coma to be correct:
> For [ItsYou.online's][iyo] user ID **,** following can be used reliably with 0-stor: email address

chrisvdg: amended

PR #289, opened 2017-11-07
ShardError.Error() now returns a pretty error message

### LINKED ISSUE #261: Zerostorcli returns an error message in case of it fails in reparring a file
**Steps:**
- I have 3 servers
- Upload a file with data=2, parity=1 and distribution size > reptation_size 
- Access those servers and delete all database files.
- Try to repair this file though zerostorcli 

**Actual result**
image

**Expected result**
This Error message should be improved. It should throw fail instead of error.

**Version**
7c623ead388511d44dcab7e3891800a42d4e4b7e

chrisvdg:
fixes #261


chrisvdg:
Error output now return something like this:

screenshot from 2017-11-07 11-14-28

iwanbk:
I think it should produces JSON or pretty JSON by default.
Because the main user of 0-stor is application software, not human.

zaibon:
That's not really true. When used with the cli, you expect the ouutput to be human readable. If you use the client programmaticly, then there is no problem cause you have the error object.

iwanbk:
> That's not really true. When used with the cli, you expect the ouutput to be human readable.

Not a big problem, we could add `PrettyError()` func

> If you use the client programmaticly, then there is no problem cause you have the error object.

The client might be different language.

> @iwanbk , woulnd't it be hard to keep the output pure json because of how Go error handling is usually done?

Yeah, then it needs to be fixed.

> Before this commit it would print json but it was prepended by a context message so it wouldn't have successfully parsed to json anyway?

Not sure, IIRC i had a case where i can parsed the json successfully.
But yeah, context message is very common.

iwanbk:
>> If you use the client programmaticly, then there is no problem cause you have the error object.

> The client might be different language.

Another issue is : the `errors` field of `ShardError` is private, not public

iwanbk:
After some more thinking, i think it is not common to have JSON string in `Error`.
So, i'm also OK with this way, but we still need to have `func` which returns error in JSON format.

iwanbk:
> After some more thinking, i think it is not common to have JSON string in Error.
> So, i'm also OK with this way, but we still need to have func which returns error in JSON format.

But `error` interface only have `Error()`.
So, i think it is still better to make JSON string as `Error()` output.

chrisvdg:
> after some more thinking, i think it is not common to have JSON string in Error.  

I was about to mention this, no one will expect that an error would return a pure JSON string, what we could do is make a ErrorJSON interface and add ErrorJSON methods to error types where we need it, and where you want JSON output you can check if it conforms to the ErrorJSON interface or something?

iwanbk:
> Another issue is : the errors field of ShardError is private, not public

nevermind, it has `Errors` func. Sorry for the noise :)

PR #290, opened 2017-11-07
Adds prettified output for cli file metadata command

### LINKED ISSUE #286: Prettyfy cli metadata output
The current output of the  `file metadata` command currently looks as follows
screenshot from 2017-11-06 15-40-19

It doesn't look very readable nor clean.
This could at the very least be turned into prettified JSON.

chrisvdg:
Adds human readable, json and prettified json output for cli metadata

Closes #286


chrisvdg:
output for the `metadata` command now looks like this:
screenshot from 2017-11-07 16-20-34
screenshot from 2017-11-07 16-22-15

chrisvdg:
I think it would be better to put them as functions in the CLI code as like you said, they don't add value to the API as they wouldn't be really used anywhere else

PR #291, opened 2017-11-08
DeleteNamespace returns error when failed

### LINKED ISSUE #247: zerostorcli doesn't return an error if it failed to delete a namespace
**Steps**
- You should have two iyo users
- As user_1 create a new namespace and send read or write invitation to user_2
- As user_2 accept this invitation
- Using user_2 iyo_id and iyo_secret, try to delete this new namespace through zerostorcli
- Check the zerostorcli output and error

**Expected result**
zerostorcli should return an error when it didnt delete the namespace.

**Version**
f71dc50492b1f2a6201b0252608aa214e19a1320

chrisvdg:
The IYO client doesn't return an error (yet) when a namespace was not successfully deleted.

As a (temporary) solution I manually check the response for a 403 which is returned when deleting the namespace failed (no permission or non existing sub org). This check can then be reverted when it's fixed in the IYO client.

closes #247


cmd/zerostorcli/namespace.go
```diff
@@ -38,9 +38,15 @@ func deleteNamespace(c *cli.Context) error {
 
 	namespace := c.Args().First()
 	if err := iyoCl.DeleteNamespace(namespace); err != nil {
-		return cli.NewExitError(err, 1)
+		errMsg := fmt.Sprintf("failed to delete namespace: %v", err)
+		if err == itsyouonline.ErrForbidden {
+			errMsg += " (please check if provided IYO app ID has the right permissions)"
```
zaibon: also explain which permission is required.

PR #295, opened 2017-11-15
Adds badger cg

### LINKED ISSUE #281: Manually trigger garbage collection of badger.
From the [badger README](https://github.com/dgraph-io/badger#garbage-collection) regarding garbage collection, now badger expect the GC to be trigger by the client.

We should define a policy to when trigger the GC and implement it in 0-stor server.

A simple approach could be to simply call the GC periodically. But we could probably come up with better strategies to try to limit the impact of the GC on the performance of the 0-stor during high load.

chrisvdg:
Adds garbage collection trigger for the Badger backend db.
For now it's time interval based.

Closes #281


chrisvdg:
One concern tho is the comment on the `PurgeOlderVersions` func from badger that mentions that this might impact long running reads.
```
// PurgeOlderVersions deletes older versions of all keys.
//
// This function could be called prior to doing garbage collection to clean up
// older versions that are no longer needed. The caller must make sure that
// there are no long-running read transactions running before this function is
// called, otherwise they will not work as expected.
```

server/db/badger/badger.go
```diff
@@ -178,3 +198,28 @@ func (b BadgerDB) List(prefix []byte) (result [][]byte, err error) {
 
 	return result, err
 }
+
+// collectGarbage runs the garbage collection for Badger backend db
+func (b BadgerDB) collectGarbage() error {
+	b.db.PurgeOlderVersions()
```
zaibon: `PurgeOlderVersions` also returns an error, I needs to be handled as well
https://godoc.org/github.com/dgraph-io/badger#DB.PurgeOlderVersions

PR #294, opened 2017-11-14
Updates docs

chrisvdg:
Updates and corrects docs

Closes #159
Closes #299

PR #302, opened 2017-11-17
fix memory DB and test Filter/List DB methods

GlenDC:
+ fixes the memory-DB as that one seemed full of bugs;
+ also adds tests for the Filter and List DB methods;
+ ensures that the DB-related tests in `db_test` are tested for both the Memory and BadgerDB implementations;
+ also updates dependencies (mostly because of the out-dated `github.com/stretchr/testify`, but the vendored badger version was also updated as it didn't match the `master` version filter in Gopkg.toml;


GlenDC:
Rebased master into this branch, as to fix the merge conflict (which happened after merging PR #294 into master).

GlenDC:
Ready for review @chrisvdg.

GlenDC:
Found some more bugs in the BadgerDB/MemoryDB code, this PR is ready for review now, @chrisvdg.

server/db/badger/badger.go
```diff
@@ -138,23 +139,28 @@ func (b BadgerDB) Exists(key []byte) (exists bool, err error) {
...
 }
 
-// Filter implements db.Filter
+// Filter implements DB.Filter
 func (b BadgerDB) Filter(prefix []byte, start int, count int) (result [][]byte, err error) {
+	if count == 0 {
+		return
+	}
+	if count == -1 {
+		count = math.MaxInt32
```
chrisvdg: Is this needed?
And also what if a user inputs `-2`? Then you allow it to run without the MaxInt32 limit?

server/db/memory/memory.go
```diff
@@ -26,12 +26,16 @@ func New() *DB {
 func (mdb *DB) Get(key []byte) ([]byte, error) {
 	mdb.mux.RLock()
 	val, exists := mdb.m[string(key)]
-	mdb.mux.RUnlock()
-
 	if !exists {
+		mdb.mux.RUnlock()
```
chrisvdg: Why not defer `mdb.mux.RUnlock()`?

server/db/badger/badger.go
GlenDC: No you're right. We should probably make that a `count < 0` check.

The reason why this is check is here, so we don't require double-checking for ever iteration such as `count > 0 && len(...) == count`. Now we can just skip that first condition.

server/db/memory/memory.go
GlenDC: defer comes at a cost. So if it can be avoided, it should be avoided.
My general rule of thumb is that a soon as code starts go too deep/complex I use defer for unlocking.

server/db/badger/badger.go
GlenDC: Can you fix it @chrisvdg?

server/db/memory/memory.go
GlenDC: See https://github.com/zero-os/0-stor/pull/302#discussion_r151977457

server/db/badger/badger.go
GlenDC: I will fix it, forgot this was my PR :')

chrisvdg: can't it be left as is, since `len()` would never return a negative number, so `len(...) == count` is always false? Just wondering

GlenDC: Sure, we can indeed leave that as it is, didn't think about that. Good catch.

Issue #68, opened 2017-07-28
client daemon


iwanbk:

zaibon:
WIP specification for the daemon: https://github.com/zero-os/0-stor/blob/master/specs/daemon.md

GlenDC:
@iwanbk can you take over this issue? I think you'll first have to finish of the daemon spec to make sure the entire design is worked out. Make sure to cover all corners and details. After that you can start implementing I suppose.

iwanbk:
> @iwanbk can you take over this issue?

sure

iwanbk:
a WIP can be found on `proxy` branch:
- partial implementation at https://github.com/zero-os/0-stor/tree/proxy/proxy
- no unit test 
- tested with simple python test file https://github.com/zero-os/0-stor/blob/proxy/proxy/python/main.py

GlenDC:
+ Looks good in general. Added the comments where needed;
+ You can probably start using `spf13/cobra` for the command-line tool library (unless you or someone else objects in #318);
+ While you should probably add some unit tests here and there, I think for the proxy it's much more important that we have a complete and decent end-to-end (system) test suite.
   + This test suite can than also serve as example code of the full API;

iwanbk:
> While you should probably add some unit tests here and there, I think for the proxy it's much more important that we have a complete and decent end-to-end (system) test suite.
> This test suite can than also serve as example code of the full API;

Indeed. i also only tested it with python script so far.
Then we might want to do #15 in the same time with this issue

GlenDC:
Yeah sure why not.

GlenDC:
You can/should use https://github.com/spf13/cobra by the way as the ctl framework

See #318 for more information.

GlenDC:
@iwanbk you'll have to rebase master into your branch and you can than add your daemon GRPC service (spawner) @ https://github.com/zero-os/0-stor/blob/master/cmd/zstor/commands/daemon.go

GlenDC:
I think https://github.com/zero-os/0-stor/blob/b6fa3b70059e8ae5e16d332f409a9472df5449a2/daemon/api/api.go#L3-L9 should be changed to the following interface:

```go
type Daemon interface {
	// Serve the Daemon on the given listener, ...
	Serve(lis net.Listener) error

	// Close closes the daemon
	Close()
}
```

[...]

Issue #312, opened 2017-11-20
proposal: remove 0-stor server


GlenDC:
I propose to remove the 0-stor server.

90% of the work is already done in the client. I'm talking about compression, encryption, chunking, erasure coding and perhaps even some more stuff.

On top of that we're also planning to provide a daemon that will run the client for you in the background (see #68). That basically turns the client into a server.

The server does really not do much at all. The only thing it does is checksum, store and fetch data. On top of that it also stores and fetches reference list. Really that is not a lot at all. It also provides permission checking using IYO JWT, but that is also almost nothing of code. To me it seems a waste of resources to provide an entirely separate server for this, while the client is already doing so much processing anyhow, it can just as well do the final part of the road and store it in the database.

So I propose we provide a nice 0-stor API (merging the client and server code into one). And on top of that provide a command line tool which allows you to run simple commands from the command line (by simply calling that 0-stor API). And lastly provide a daemon server which exposes the API as a GRPC service.

GlenDC:
As our main language language is Golang, I would keep the reference implementation in Golang. 

However it might be interesting to also provide a Rust version. Shouldn't take too long, as 0-stor is relatively simple. Obviously we would first want to prototype to ensure that we indeed get a big performance boost. But if that is the case, it would be nice to provide this alternative Rust version. Because if we're anyway going to rely on a daemon service for clients such as the Python client, than it doesn't really matter any longer if that daemon is written in Rust or Golang. On top of that it gives us the performance that we can get with a language like C, while allowing us to do so in a sane language environment.

GlenDC:
This is just a proposal though, happy and very curious to hear your opinions on this one @zaibon.

iwanbk:
I think i really missed something here.
If we don't have the server, how we store the data on our 0-stor server?

> And lastly provide a daemon server which exposes the API as a GRPC service.

This is exactly what the current server does, no?

GlenDC:
> I think i really missed something here.

Nothing missed, I'm proposing it out of the blue

> If we don't have the server, how we store the data on our 0-stor server?

You wouldn't, you would store it directly in the Database. It's not like we now store something in the 0-stor server, the 0-stor server stores content in the Database as well. Well to be fair you wouldn't store it in the Database yourself, that would happen for you by using the API (client), just as you kinda do now. It's just that we're cutting out another connection (client->server).

> This is exactly what the current server does, no?

[...]

PR #315, opened 2017-11-20
Update daemon spec

iwanbk:
Part of #68


zaibon:
Let's wait #312 is resolved before we merge this, since depending on #312 we could change a bit the organization of this task.

GlenDC:
> Let's wait #312 is resolved before we merge this, since depending on #312 we could change a bit the organization of this task.

The proposal given in issue #312 has been declined, and thus has no impact on this PR any longer.

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
 ### future feature
 In a second phase we want to add a caching and autodiscovery feature to the deamon so it could be used as some kind of IPFS replacement.
 
-
 ## Requirements:
 - The daemon should be started from the client CLI. We pass the same configuration as we would for a normal client.
```
GlenDC: The daemon should be a separate binary:

+ Either it can be just that a daemon;
+ Or we have 2 binaries (a server and a daemon), so that you can start this service either as a foreground or background process (and also provide the option to stop that background process);
+ Or we have one binary service binary that can be started either as a foreground process (server) or background process (daemon);

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
 - Simple interface, probably grpc. Grpc has a great number of supported language, that should give us the flexibilty we need.
-- Commands:
-  - SetObject(key, path)
-  - GetObject(key, path)
+- there is no authentication needed between daemon user and the daemon
+- IYO commands:
```
GlenDC: Why do we need all these commands, when really 0-stor should be just about storage of objects? I thought the management of IYO and namespaces was done outside this project?

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
+  - CreateJWT(namespace, permission)
+  - CreateNamespace(namespace)
+  - DeleteNamespace(namespace)
+  - GivePermission(namespace, userID, permission)
+  - RemovePermission(namespace, userID, permission)
+  - GetPermission(namespace, userID)
+  
+- Object Commands:
+  - Write(key/meta, value, referenceList, optional prevKey/prevMeta): write value to given key/meta with specified reference list. If optional prevKey/prevMeta is specified, we build metadata linked list
```
GlenDC: `with optionally a specified reference list.`*

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
+  - WriteStream(key/meta, value stream, referenceList, optional prevKey/prevMeta): same as `Write` but receive stream of value instead of value
+  - Read(key/meta) (value, referenceList): read object with given key/meta
+  - ReadStream(key/meta) (value stream): Same as `Read` but returns stream of value
+  - Delete(key/meta): Delete object with given key/meta
+  
+- Walk commands
+  - Walk(startKey, startEpoch, endEpoch) (stream of object): Walk over the metadata.
```
GlenDC: Why do we think in terms of epoch here? I don't get it.

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
 In a second phase we want to add a caching and autodiscovery feature to the deamon so it could be used as some kind of IPFS replacement.
 
-
 ## Requirements:
 - The daemon should be started from the client CLI. We pass the same configuration as we would for a normal client.
 - Simple interface, probably grpc. Grpc has a great number of supported language, that should give us the flexibilty we need.
-- Commands:
-  - SetObject(key, path)
-  - GetObject(key, path)
+- there is no authentication needed between daemon user and the daemon
```
GlenDC: > Because the authentication is done in the server side.

Might be useful to mention that here. As I suppose that's the reason. That and the fact that the daemon is supposed to run locally and thus the user can be trusted.

specs/daemon.md
```diff
@@ -8,10 +8,33 @@ We do this to prevent to have to re-implement all the client logic in every lang
 ### future feature
 In a second phase we want to add a caching and autodiscovery feature to the deamon so it could be used as some kind of IPFS replacement.
```
GlenDC: I would really move this out of this spec file into a separate spec, as now it just distracts without giving any more context and information.

specs/daemon.md
zaibon: These command are also present in the client API. Cause we want the user to be able to handle all the management of the zero-stor from a single interface.
I find it cumbersome for a user to have to use something else then the 0stor client before beeing able to actually send anything to the server.

specs/daemon.md
zaibon: Regarding writes I wonder if we should not also allow a write method that get a file path as data source. so the daemon could actually read the data directly from the filesystem. This would remove the need of sending the data through the grpc between the python client and the daemon.

zaibon:
As a general comment, I wonder if we shouldn't rename this component from daemon to proxy. cause at then end it's really what it is, just a proxy between a client written in another language that go and the server.

specs/daemon.md
GlenDC: Ok Fair enough.

specs/daemon.md
GlenDC: Maybe that could just be a 3rd write command, so we would have `Write`, `WriteStream` and `WriteFile`?

GlenDC:
> As a general comment, I wonder if we shouldn't rename this component from daemon to proxy. cause at then end it's really what it is, just a proxy between a client written in another language that go and the server.

Yeah I think proxy is an acceptable name. If so I propose we have the following cmds:

- `ztorctl` (client command line tool);
- `ztordb` (database server);
- `ztorproxy` (server/daemon that provides a GRPC interface for the client);

specs/daemon.md
iwanbk: agree as separate binary

specs/daemon.md
iwanbk: yep

specs/daemon.md
iwanbk: not sure i get your comment.

We use epoch because we don't always need to walk over all metadata.
But i think it is not what you asked.

specs/daemon.md
iwanbk: agree with `WriteFile`

GlenDC: Can you please also provide one or multiple examples for each object command, @iwanbk? This will help motivate their existence.

specs/daemon.md
GlenDC: No it's not. But don't worry, It's probably because I don't understand the 0-stor client code and API well enough yet. Will start to dive into the 0-stor client code tomorrow.

iwanbk:
@GlenDC i addressed your review

specs/daemon.md
```diff
@@ -8,10 +8,36 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
+  - DeleteNamespace(namespace)
+  - GivePermission(namespace, userID, permission)
+  - RemovePermission(namespace, userID, permission)
+  - GetPermission(namespace, userID)
+  
+- Object Commands:
+  - Write(key/meta, value, referenceList, optional prevKey/prevMeta): write value to given key/meta with optionally specified reference list. If optional prevKey/prevMeta is specified, we build metadata linked list.
+    We use this command to write a not so big data. If data is considered big, we should use `WriteStream` or `WriteFile`
+    
+  - WriteFile(key/meta, file path, referenceList, optional prevKey/prevMeta): same as `Write` but receive file path instead of value. We use this command to store a file to 0-stor server.
```
zaibon: The point of this command is that the proxy will read the file directly instead of passing through the grpc. So it should be faster. Also we need to specify that the file need to be readable from the proxy for this call to work.

specs/daemon.md
```diff
@@ -8,10 +8,36 @@ We do this to prevent to have to re-implement all the client logic in every lang
...
+  - WriteStream(key/meta, value stream, referenceList, optional prevKey/prevMeta): same as `Write` but receive stream of value instead of value. We use this command when we have stream of data which could be from external source like network or any reader. 
+  - Read(key/meta) (value, referenceList): read object with given key/meta. We use this command when we expect the value to be not so big.
```
zaibon: Shouldn't we also have a ReadFile to be symmetric with the write api? Not sure about this one though.

specs/daemon.md
iwanbk: added

specs/daemon.md
iwanbk: added

iwanbk:
@zaibon i addressed your review, rename `daemon` to `proxy`, and also renamed the file to `proxy.md`

specs/daemon.md
GlenDC: I finally understand why. It's because of the metadata and how we store it. After going through the client code I now understand why the Walk command is defined as it is.

PR #329, opened 2017-11-21
fix all warnings and formatting in (golang) codebase

### LINKED ISSUE #303: fix all warnings and formatting in (golang) codebase
Warnings should be treated as errors, and be fixed. Using the Golang toolchain you can easily check for and list warnings if they exist:

```
go vet $(go list ./... | grep -v vendor)
```

You can format all code at once using:

```
go fmt $(go list ./... | grep -v vendor)
```

For now it seems however that all formatting is fine as it is on the master branch.

Note that you should install, configure and use Golang plugins in your favorite editor/IDE, such that the warnings are automatically listed and formatting is automatically corrected when saving a file.

Once all warnings have been fixed (in a reasonable way), also ensure to add the following to our `.travis.yml` file, such that we also have a Travis CI test that ensures all checked in code is warning free and correctly formatted:

```diff
diff --git a/.travis.yml b/.travis.yml
index fcd3a6b..6d51112 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,13 @@ matrix:
       script: make
 
     - language: go
+      go: 1.9
+      install: skip
+      script:
+        - go vet $(go list ./... | grep -v vendor)
+        - diff -u <(echo -n) <(gofmt -d $(find . -name "*\.go" | grep -v '\.git/' | grep -v vendor))
+
+    - language: go
       sudo: true
       dist: trusty
       go: 1.9
```

katia-e:
Fixes #303.


katia-e:
Now new Error interface formats error output similar way as formatPrettyError(), but for single ServerError. In this case, should formatPrettyError() maybe also use ServerError.Error() interface in order to create error message for a slice of []ServerError?

GlenDC:
> Now new Error interface formats error output similar way as formatPrettyError(), but for single ServerError. In this case, should formatPrettyError() maybe also use ServerError.Error() interface in order to create error message for a slice of []ServerError?

Will check it out and fix it in another commit if needed, might be easier than to explain.

> Haven't added .travis.yml yet. @GlenDC , could you give me a bit of insight what is this about?

Travis is a Continuous Integration (CI) tool, which allows for OSS projects to use it for free, for all kind of things, going from running unit tests to shipping code.

For a project to make use of it, it has to be enabled by someone with admin powers over the repo on http://travis-ci.org. I think it might work without, but after that you would normally also add a `.travis.yml` configuration file in the root directory of your repo.

In the case of 0-stor (and all zero-os repositories AFAIK), this is already done for you. So if you go to our repo's root dir you'll find our configuration file here: https://github.com/zero-os/0-stor/blob/master/.travis.yml

If you want to learn to understand that configuration file (as it really is not that difficult once you get to know it), you can head over to: https://docs.travis-ci.com

But you do not really need to know each and every bit of the travis config file of our project, in order to add a test to it. In this case if you already know that each block such as https://github.com/zero-os/0-stor/blob/35631819c202472f6ef629a84a6bd7debee8c02d/.travis.yml#L20-L25 will run as a separate test on some VM, than that's already fine for this purpose.

So all you need to do now, is copy such a block, paste it and modify it. You'll want to make sure that the go version is `1.9`, we don't require sudo, we can skip installation and for the scripts you would add the following:

```yaml
script:
    - go vet $(go list ./... | grep -v vendor)
    - diff -u <(echo -n) <(gofmt -d $(find . -name "*\.go" | grep -v '\.git/' | grep -v vendor))
```

Which will make it run this block, to ensure that all warnings are fixed AND that all code is properly formatted.

These Travis CI tests (each block is a test) is run for each commit that you push to GitHub. If the commit is part of a branch which is part of an open PR attached to this repository, it will run each block two times (which is why you see 2 checks instead of one). One check (group of tests) is with the code as it is on your branch. The second test is with the code already merged together with the target branch, to (try to) ensure that after the merge the test still passes.

GlenDC:
I added it to travis myself, and also did some last quick improvement to shard_error. So that answers your 2 questions in the form of 2 commits. Please make sure you understand all changes in those commits @katia-e and the motivations behind those changes. Feel free to link here on in telegram any questions you might have, if there are any.

PR #337, opened 2017-11-22
rename server/disk package to server/fs and improve

### LINKED ISSUE #323: rename server/disk package to server/fs and improve
More info: https://github.com/zero-os/0-stor/blob/master/specs/server.md#fs

katia-e:
fixes #323


server/fs/space.go
```diff
@@ -1,4 +1,8 @@
-package disk
+/*Package fs defines a package to estimate available space in 0-stor
+
+FreeSpace is a function that returns available space in bytes
```
chrisvdg: This can be omitted as it is already stated above `FreeSpace` what it does

Also in case other operations on the file system need to be added, they will be added in this package and the package description wouldn't make sense anymore
I guess you can make the package godoc as simple as:
```go
//Package fs defines a package that handles file system interactions on the 0-stor server
package fs
```

Issue #327, opened 2017-11-21
Move all GRPC code to the server/api/grpc package


GlenDC:
+ [x] move all GRPC-related code to the `server/api/grpc` package;
+ [x] ensure we define the `grpc.API` object which implements the `api.API` interface;
+ [x] improve and adapt the code where needed;
+ [x] make sure it's all properly documented;
+ [x] make sure it's all properly unit tested where required;
+ [x] ensure unit tests cover at least 80% of the code;

More information: https://github.com/zero-os/0-stor/blob/master/specs/server.md#api

Issue #317, opened 2017-11-20
create GRPC middleware to verify JWT


GlenDC:
Using the JWT Verifier added as the solution for issue #310, we should create a GRPC middleware in the GRPC server code as to verify JWTs like that, rather than having it to do automatically as we currently do using code such as:

> https://github.com/zero-os/0-stor/blob/0b38327dabe35ee8a08dffed117fb035b336b0b9/server/object_api.go#L32-L34

Instead we should write and use a GRPC middleware that does this for us automatically. This keeps the logic also in one single location.

Some pointer about grpc middleware :
- https://github.com/grpc-ecosystem/go-grpc-middleware
- https://github.com/grpc/grpc-go/blob/master/interceptor.go

GlenDC:
@chrisvdg once you're finished with #310, you can start one this one.

PR #340, opened 2017-11-22
Reworks servers's JWT handling

chrisvdg:
* Reworks servers's JWT handling
* Adds testing for JWT


server/jwt/verifier.go
```diff
@@ -0,0 +1,289 @@
...
+
+// jwtCacheVal represents the value stored in the cache
+type jwtCacheVal struct {
+	valid  bool
+	scopes []string
+}
+
+// NewDefaultVerifier returns a JWT verifier
+// with the IYO(https://itsyou.online/) public key
+func NewDefaultVerifier(authEnabled bool) *Verifier {
```
GlenDC: This function should take no parameters.
Also rename it to `DefaultVerifier`, as we don't always create a `New` one, only the first time.

server/jwt/verifier.go
```diff
@@ -0,0 +1,289 @@
...
+	createDefVerifierOnce.Do(func() {
+		var err error
+		defVerifier, err = NewVerifier(iyoPublicKeyStr, authEnabled)
```
GlenDC: should change to `NewVerifier(iyoPublicKeyStr, true)`

server/jwt/verifier.go
```diff
@@ -0,0 +1,289 @@
...
+
+// GetScopes returns the scopes from a JWT
+func (v *Verifier) GetScopes(token string) ([]string, error) {
+	scopes, inCache, err := v.getScopesFromCache(token)
+	if err != nil {
+		// there is error in token
+		return nil, err
+	}
+	if inCache {
+		// token is not in cache
```
zaibon: shouldn't it be `// token is in cache` since we check that `inCache` is true?

server/jwt/verifier.go
zaibon: So if we don't pass any argument there. How can we create a test verifier ?
Shouldn't we have a jwtVerifier interface then, that can be implement by a normal verifier and a testing verifier ?

GlenDC: You wouldn't use the DefaultVerifier for testing purposes... You would create your own verifier using `NewVerifier`and your own test key... Just as you do already on master...

zaibon: ha yeah, didn't pay attention this was part of the default one. make sense.

GlenDC: Perhaps Default is not a good name, but it's the best I could come up with. The reason this global one exists, is so that the cache can be shared for all those that wish to use that key. Whose key is that actually? I don't get how we can hard-code a key for IYO? Shouldn't that be a key coming from the 0-stor library user @zaibon?

server/jwt/verifier.go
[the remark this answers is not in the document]
GlenDC: Jup seems like it.

On a side note, I think it's a bit overkill to return 3 parameters for this though.
As it's an internal API and the result is a slice (so no danger of nil-pointer-deref) I think it's fine to just return an empty slice (`nil`) in case it wasn't in the cache.

This way you can just check `if scopes != nil`. Just my 2 cents, not too important, but I think it's a bad sign when you have more than 2 return parameters.

server/test/grpc_object_test.go
```diff
@@ -16,12 +16,17 @@ import (
...
-	server, iyoCl, org, clean := getTestGRPCServer(t)
+const (
+	// authentication key where the JWT can be found in the GRPC's context
+	authGRPCKey = "authorization"
+)
 
-	namespace := fmt.Sprintf("%s_0stor_testnamespace", org)
-	bufList := populateDB(t, namespace, server.DB())
+func TestListObject(t *testing.T) {
+	organization := "testorg"
```
GlenDC: My dear @chrisvdg, can you please start using constants when they aren't supposed to be variables... It's not the first time I tell you this. Always default to constants if possible. And it's not just here, it seems it's the same issue in the entire file.

server/object_api.go
```diff
@@ -7,19 +7,22 @@ import (
...
 
 type ObjectAPI struct {
-	db db.DB
+	db          db.DB
+	jwtVerifier *jwt.Verifier
 }
 
-func NewObjectAPI(db db.DB) *ObjectAPI {
+func NewObjectAPI(db db.DB, v *jwt.Verifier) *ObjectAPI {
 	return &ObjectAPI{
```
GlenDC: Can you add a panic please if either `db` or `v` is `nil`. This will make it panic at initialization rather than somewhere later at first usage.

server/namespace_test.go
```diff
@@ -29,8 +30,7 @@ func getTestNamespaceAPI(t *testing.T) (*NamespaceAPI, func()) {
 		os.RemoveAll(tmpDir)
 	}
 
-	disableAuth()
-	api := NewNamespaceAPI(db)
+	api := NewNamespaceAPI(db, jwt.NewDefaultVerifier(false))
```
GlenDC: Should be:

```go
pubKey, err := ioutil.ReadFile("../../devcert/jwt_pub.pem")
require.NoError(err)
verifier, err := jwt.NewVerifier(pubKey, false)
require.NoError(err)
api := NewNamespaceAPI(db, verifier)
```

server/namespace_api.go
```diff
@@ -5,26 +5,29 @@ import (
...
 
 type NamespaceAPI struct {
-	db db.DB
+	db          db.DB
+	jwtVerifier *jwt.Verifier
 }
 
-func NewNamespaceAPI(db db.DB) *NamespaceAPI {
+func NewNamespaceAPI(db db.DB, v *jwt.Verifier) *NamespaceAPI {
 	return &NamespaceAPI{
```
GlenDC: Can you please panic if either `db` or `v` is `nil`. This way we get a panic at initialization rather than somewhere later during the first usage.

server/jwt/verifier.go
zaibon: > Shouldn't that be a key coming from the 0-stor library user @zaibon?

This key is the public key IYO use to sign the JWT that it generates. it could come from the user, but so far there is no plan for this key to be changed.

https://github.com/itsyouonline/identityserver/blob/01863b33a062d98ef9580829f2e44418b213012f/docs/oauth2/jwt.md#case-1-use-an-oauth2-token-for-jwt-creation-where-the-jwts-claim-set-is-a-subset-of-the-oauth-tokens-scopes

What we could do is let the user provide a key, but still have a default. Cause forcing any user to go read the doc to know which public key he need to use to validate IYO JWT seems a bit out of scope of the 0-stor and cumbersome

server/test/grpc_object_test.go
zaibon: most probably my fault to not use constant where I can on the rest of this file.

server/jwt/verifier.go
chrisvdg: should `AuthDisabled` also be removed from the config settings? Since authentication will always be enabled now with the `DefaultVerifier`

GlenDC: > What we could do is let the user provide a key, but still have a default.

Seems like a good idea 👍

> should `AuthDisabled` also be removed from the config settings? Since authentication will always be enabled now with the `DefaultVerifier`

I suppose you ask because we might want to start a 0-stor server for dev purposes, where we don't want to verify either. Not sure what to do with that one for now. Need to think about it, or perhaps @zaibon has an idea already.

zaibon: I think we can even advertise this has a feature. the possibility to start the 0-stor without the hassle of having to authenticate to access a namespace. I can see cases where this can be used.

GlenDC: Perhaps @zaibon. But to me it would make more sense than to make the entire JWT verifier optional. There is no point in going through all that logic, if we do not care about validation anyhow. Like the only public function we seem to be using is `ValidateJWT` as far as I can see in the server code.

So once @chrisvdg created the middleware in issue #317, he should ensure that this middleware becomes optional, and is only to be used when we actually want to use JWT and verify tokens/scopes. For the 0-stor server we could than provide a flag to disable this JWT verify (could be as simple as `--no-jwt-verify`).

If we go down this road, we can already remove the `authEnabled` property from the Verifier. We could than create a `NopVerifier` on top of this, which simply returns `nil` for the `ValidateJWT` method. This might be useful for the current setup and test code.

In the future however, I think it would be useful when we create the `grpc.API` (as described in open issue #327), that:

+ we can give it a custom pubKey (would use NewVerifier instead of DefaultVerifier);
+ we can make JWT verification entire optional (wouldn't create the JWT middleware at all);

In this case we wouldn't really need the `NopVerifier` any longer in production code.

Also as we only ever use the `ValidateJWT` we should perhaps make all the other methods private, as I don't see why we should expose those, even though we don't ever need them (and we're not even using any of those outside the jwt package).

GlenDC:
All in all, looked good @chrisvdg. One last thing. You should make your constructors declare the real types, not an interface. I think it's OK to define an interface in the package itself. But you should also expose the real type (unless you have a very good reason not to).

Issue #332, opened 2017-11-21
rename client/lib package to client/components


GlenDC:
Package names are supposed to be meaningful. In golang every package is a lib (as in library), so that name is not really useful at all. As this package (and especially its subpackages) seem to be defining components, I think it would be useful if we simply rename that package to `github.com/zero-os/0-stor/client/components` instead of the current `github.com/zero-os/0-stor/client/lib`.

PR #349, opened 2017-11-24
replace client lib READMEs with Godocs #331

### LINKED ISSUE #331: replace client lib READMEs with Godocs
For some reason all lib components now have a markdown README.me file.

The documentation part of it should be simply added to the main file (e.g. `chunker.go`), or a `doc.go` file if you really want that, but I wouldn't really do that in the case of these small packages. This kind of package documentation has to be added to the top of the file (above the `package` statement).

As an example, that would mean something along the following lines for the chunker package:

```diff
diff --git a/client/lib/chunker/chunker.go b/client/lib/chunker/chunker.go
index 3490458..3cf7b46 100644
--- a/client/lib/chunker/chunker.go
+++ b/client/lib/chunker/chunker.go
@@ -1,3 +1,8 @@
+// Package chunker defines a pipeline component
+// to split a block into multiple smaller blocks.
+//
+// Chunker returns an iterator that will yield a new chunk of data of the choosen size.
+// The block can then be sent to the rest of the pipeline.
 package chunker
 
 import (
```

The example written in the README.md file should be added into a new file (one for each component package), named `example_test.go`. This way the example will be both added to the package Godocs, as well as be run alongside the other unit tests. This ensures that the example stays in sync with the code.

By doing these changes we:

+ Keep the documentation in reach of the programmer:
  + The programmer can read the docs directly alongside the code;
  + The documentation will be available in its natural habitat on Godocs;
+ Ensure that the examples are correct, will stay in sync with the code and can be copy pasted;

The following component packages have currently such a README.md file that will have to be transformed to Godocs documentation and runnable example tests:

+ [ ] `github.com/zero-os/0-stor/client/lib/chunker`:
         https://github.com/zero-os/0-stor/tree/master/client/lib/chunker/README.md;
+ [ ] `github.com/zero-os/0-stor/client/lib/compress`:
          https://github.com/zero-os/0-stor/tree/master/client/lib/compress/README.md;
+ [ ] `github.com/zero-os/0-stor/client/lib/distribution`:
         https://github.com/zero-os/0-stor/tree/master/client/lib/distribution/README.md;
+ [ ] `github.com/zero-os/0-stor/client/lib/encrypt`:
          https://github.com/zero-os/0-stor/tree/master/client/lib/encrypt/README.md;
+ [ ] `github.com/zero-os/0-stor/client/lib`:
          https://github.com/zero-os/0-stor/tree/master/client/lib/README.md;

While transforming those examples and documentation:

+ Ensure to improve/fix the documentation as you go. Don't be shy to expand the documentation, as I'm sure it can use some improvement;
+ When writing documentation, at the very least also run it through a English (US) Spellchecker, this never hurts;
+ Ensure to improve/fix any mistakes in the examples;
+ If you think some examples are missing, add them as well;

You don't have to publish your code to the github master branch to see your Godocs rendered. You can run godocs locally by:

1. run godoc on port `6060` (or any free local port): `$ godoc -http=:6060`;
2. open a browser and surf to: http://localhost:6060/pkg/github.com/zero-os/0-stor/client/lib/#pkg-subdirectories
3. enjoy reading your rendered documentation and seeing your examples

Pro tip: you can modify your godocs code (examples/documentation), save the file and simply refresh the web page. It will automatically re-render your update.

Here are some links to some related golang documentation:

+ https://blog.golang.org/godoc-documenting-go-code
+ https://blog.golang.org/examples

katia-e:
Fixes #331


client/lib/encrypt/encrypt.go
```diff
@@ -1,3 +1,9 @@
+/*Package encrypt defines a package that handles encryption/decryption of the input data
+
+Package Supports Asymmetric & Symmetric algorithms.
```
GlenDC: You can remove this line, as it is not true. The package currently only supports AES, which is symmetric. Obviously nothing would stop us from supporting an encryption/decryption component that uses asymmetric algorithms, but for now we don't. So let's not put false statements in our documentation.

client/lib/encrypt/encrypt.go
```diff
@@ -1,3 +1,9 @@
...
+Currently Supported encryption algorithms:
+    	- AES GCM
```
GlenDC: I would put it as:

```
- AES256 (using GCM);
```

As that's more complete and correct with what the code reflects.

client/lib/encrypt/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/zero-os/0-stor/client/lib/encrypt"
+)
+
+func Example() {
+	// given data ...
+	data := []byte("hello world")
+
+	// given private key and nonce ...
+	keySize := 32
```
GlenDC: constants...

client/lib/encrypt/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	nonceLen := 12
```
GlenDC: The nonce is not used and thus shouldn't be declared.

A nonce is declared automatically and for every Encrypt call. It is part of the output data.

client/lib/encrypt/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	privKey := make([]byte, keySize)
```
GlenDC: For an example I wouldn't generate a random private key.

An example needs to reflect a real use-case more or less,
so let's just use some memorable password using random words, such as:

```
france-though-athens-every-force
```

client/lib/encrypt/example_test.go
GlenDC: Than again, just delete these, as we don't need them if you follow my other comments.

GlenDC:
> Shouldn't we rename the lib package to components ?

There is a seperate issue for it open (issue #332), since a while ago.

client/lib/compress/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/zero-os/0-stor/client/lib/compress"
+	"github.com/zero-os/0-stor/client/meta"
+
+	"github.com/zero-os/0-stor/client/lib/block"
+)
+
+func Example() {
+	// given payload ...
+	payload := make([]byte, 4*4)
```
GlenDC: Can you please define some static non-random/sequential input. Just make a payload of a random sentence. At least that would show compression happened, now it looks like nothing happened.

client/lib/encrypt/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	fmt.Printf("Initial data: %v\n", string(data))
+	fmt.Printf("Result of encryption/decryption: %v\n", string(decrypted))
+	if bytes.Compare(data, decrypted) != 0 {
+		log.Fatalf("decryption failed")
+	}
+	fmt.Println("Decrypted data matches initial data")
+	// Output:
+	// Initial data: hello world
+	// Result of encryption/decryption: hello world
```
GlenDC: I would change this to `encryption+decryption` (`+` instead of`/`).

client/lib/example_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+
+	"github.com/zero-os/0-stor/client/lib/block"
+	"github.com/zero-os/0-stor/client/lib/compress"
+	"github.com/zero-os/0-stor/client/meta"
+)
+
+func Example() {
```
GlenDC: I would remove this file, it doesn't really show me anything useful.

client/lib/compress/example_test.go
katia-e: @GlenDC , 
>...define some static non-random.., 
> ...make a payload of a random sentence, 

I didn't understand, whether payload should be random or not in the end?

> ...now it looks like nothing happened
How exactly do I show that compression happened? Output compression result?

GlenDC: > I didn't understand, should payload be random or not in the end?

It shouldn't be random. It should be some fixed string, sorry for my ambiguous use of random. I just meant to say, choose any sentence, and hardcode it to be used for this example.

GlenDC: > How exactly do I show that compression happened? Output compression result?

You already show it, it's just that you can't compress a [0...15] sequence, as all bytes are unique, so not much you can do to make it compressed. In fact, in some compression algorithms, for a sequence like that, your output might even be larger than your input.

client/lib/compress/example_test.go
```diff
@@ -37,11 +34,11 @@ func Example() {
...
-	fmt.Printf("After compression-decompression:\n%v\n", decompressed)
+	fmt.Printf("Payload:\n%v\n", string(payload))
+	fmt.Printf("After compression-decompression:\n%v\n", string(decompressed))
 	// Output:
 	// Payload:
-	// [0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15]
+	// No lemon, no melon
 	// After compression-decompression:
-	// [0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15]
+	// No lemon, no melon
```
GlenDC: Still no compression at all. No worries, will see if I can get it to compress one sentence, because now  the example just looks silly. But I'll do that before merging, not to worry, it's not important.

PR #362, opened 2017-11-28
add missing unit tests for client's ShardError

### LINKED ISSUE #326: add missing unit tests for client's ShardError
Add tests for lib package defined in  0-stor/client/lib/shard_error.go

+ One simple test to ensure that `Nil`/`Num`/`String` works in all scenarios (no error, one error, multiple errors). These can all be tested in one single unit test function. for `String` you shouldn't compare an exact string, instead just ensuring it's empty when it should be empty and it's not empty when it's not, should be fine enough for that method;
+ One test to ensure that ShardError is thread-safe.

katia-e:
closes #326


client/components/shard_error_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestShardError(t *testing.T) {
+	var shardErr ShardError
+
+	require.True(t, shardErr.Nil(), "must return true if no errors")
+
+	assert.Equal(t, shardErr.Num(), 0, "incorrect number of errors")
```
GlenDC: You can turn this into `require.Equal`.

Also if you do multiple requires, you can create a require object (`require := require.New(t)`), this way you don't have to pass in the test `T` object yourself.

client/components/shard_error_test.go
```diff
@@ -0,0 +1,50 @@
...
+	const n = 1000
+	var shardErr ShardError
+	var wg sync.WaitGroup
+
+	// create n threads to add errors to shardErr
+	for i := 0; i < n; i++ {
+		wg.Add(1)
+		go func() {
+			shardErr.Add([]string{"22379"}, "", nil, 0)
+			defer wg.Done()
```
GlenDC: Either you don't defer and do a plain function call, or you do a defer at the top of your function, but what you're doing now is a bit silly.

PR #363, opened 2017-11-28
Adds use of middleware

### LINKED ISSUE #317: create GRPC middleware to verify JWT

### LINKED ISSUE #346: create GRPC middleware to gather requests stats
In the same vein as #317 we should also move the gathering of statistics into a middleware. 
Current in each grpc handler we do something like https://github.com/zero-os/0-stor/blob/65060292f43a0c20a356f8c8b1e8004a14ba859f/server/object_api.go#L30

This should be moved out of the handler and moved into a middleware.

The only thing I'm not sure about his how we can  deduct which type of request (read/write) it is from the middleware.

chrisvdg:
Places following logic in grpc middleware

- [x] authentication
    - [x] stream
    - [x] unary

- [x] stats
    - [x] stream
    - [x] unary

Closes #317
Closes #346


chrisvdg:
Blocker atm is being able to get the label from the stream in the stream interceptor

server/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+	case "/NamespaceManager/Get":
+		return methodAdmin, nil
+
+	default:
+		return methodInvalid, status.Error(codes.Unauthenticated, "method not recognized by authentication middleware: "+grpcMethod)
+	}
+}
+
+// GettableLabel represents an interface that returns a label
+type gettableLabel interface {
```
GlenDC: please do not define an interface for it, there is no reason for that. Just do a cast directly using the single-method-interface you care about, where you need it.

server/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+// slice mapping to the individual method strings
+var _MethodMap = map[method]string{
+	methodWrite:   _methodStrings[0:5],
+	methodRead:    _methodStrings[5:9],
+	methodDelete:  _methodStrings[9:15],
+	methodAdmin:   _methodStrings[15:20],
+	methodInvalid: _methodStrings[20:27],
+}
+
+func getMethod(grpcMethod string) (method, error) {
```
GlenDC: I wouldn't do it like this, I would  create an actual mapping, is going to be way faster than doing an entire switch case thing. Also please remove the namespace first (but I guess make sure first that it is the correct namespace), such that you have to just check the string `List` instead of `/ObjectManager/List`.

But than I would really want to see just a mapping.

server/middleware_auth.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *grpcServer) jwtAuthenticator(ctx context.Context, grpcMethod, label string) (context.Context, error) {
+	method, err := getMethod(grpcMethod)
+	if err != nil {
+		return ctx, err
+	}
+
+	var jwtMethod jwt.Method
+	switch method {
+	case methodRead:
```
GlenDC: Are you kidding me? You are dong a double mapping here? Just map the method directly to the actual jwt method...

GlenDC:
I think for this to work properly we'll want to have the label to be moved to some metadata value (also I think label is not a good name, sounds really confusing to me). I think this is what the metadata is made for, rather than the hacky thing what we do currently (casting the request to some random interface we care about).

Some example(s) on how this can be done using metadat can be found in https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/auth

zaibon:
Using the metadata is also the trick we got on gitter yesterday to solve our problem. 
I still forgot that in grpc we control both end of the requests, so yeah there is no restriction and we can be sure to have all the metadata properly set by the client.

server/middleware_auth.go
```diff
@@ -1,83 +1,50 @@
...
-	pb "github.com/zero-os/0-stor/server/schema"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 // AuthFunc represents an authentication function
-type AuthFunc func(ctx context.Context, grpcMethod, label string) (context.Context, error)
+type AuthFunc func(ctx context.Context, grpcMethod string) error
```
zaibon: Any reason we expose this type ?

chrisvdg: Nope, fixed it

server/middleware.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	namespaceMap = map[string]jwt.Method{
+		"Get": jwt.MethodAdmin,
+	}
+)
+
+func getJWTMethod(grpcMethod string) (jwt.Method, error) {
+	// check namespace
+	switch {
+	case strings.HasPrefix(grpcMethod, objectPrefix):
```
GlenDC: You don't need to use the `strings.Replace`. You can simply skip the prefix, as you know it has the prefix. See https://play.golang.org/p/hAky8mlhkL

server/middleware_stats.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func statsLogger(ctx context.Context, grpcMethod string) {
+	label, err := extractLabel(ctx)
+	if err != nil {
+		log.Errorf("Stat was not logged due to error: %v", err)
+	}
+
+	method, err := getJWTMethod(grpcMethod)
```
GlenDC: I would make a seperate mapping and util func to get the value of that mapping, such that you can do

```go
statsFunc, err := getStatsFunc(grpcMethod)
// ...
statsFunc(label)
```

This is on top of that also required for the NameSpace.Get method, which we do want to log as a Read stats, as we do read, while for JWT that's considered admin.

client/stor/grpc.go
```diff
@@ -188,15 +188,15 @@ func (c *client) ObjectsCheck(ids [][]byte) ([]*pb.CheckResponse, error) {
...
 		Label:         c.namespace,
 		Key:           id,
 		ReferenceList: refList,
 	})
 
 	return err
 }
-func ctxWithJWT(jwt string) context.Context {
-	md := metadata.Pairs("authorization", jwt)
+func ctxWithJWT(jwt, label string) context.Context {
```
GlenDC: rename to `contextWithMetadata`.

server/middleware_auth.go
```diff
@@ -0,0 +1,48 @@
...
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+// authFunc represents an authentication function
+type authFunc func(ctx context.Context, grpcMethod string) error
+
+func unaryJWTAuthInterceptor(authFunc authFunc) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		err := authFunc(ctx, info.FullMethod)
```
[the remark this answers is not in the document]
zaibon: you mean `jwtAuthenticator` instead of `statsLogger` right ?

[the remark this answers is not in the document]
GlenDC: Yes I do.

GlenDC:
Thank you for taking the time to do these changes @chrisvdg.

However, please don't rush and do a job half-done with some silly mistakes. Either do a job properly or just say you don't have time for it. But don't go for the middle road, that doesn't help anyone.

Honestly I don't mind fixing these remarks myself, it's done quickly enough. So next time when you don't have the time to do so, just let me know. Or stay a bit longer, and finish a job well done.

PR #370, opened 2017-11-29
replace --bind with --listen flag for zstordb

zaibon:


cmd/flag.go
```diff
@@ -35,3 +36,35 @@ func (s *Strings) Set(str string) error {
...
+// ListenAddress is a string representing a host and a port
+// which can be used as a flag for a Cobra command.
+type ListenAddress string
+
+// String implements spf13/pflag.Value.String
+func (b *ListenAddress) String() string {
+	if len(*b) == 0 {
+		return ":8080"
+	}
+	return string(*b)
```
GlenDC: you can just do `return *b` in cases like this, no explicit type cast is needed, if you return it.

cmd/zstordb/commands/root.go
```diff
@@ -140,9 +140,14 @@ func ensureStoreStat(kv db.DB, dataPath string) error {
...
 
+// func validateListen(hostport string) error {
+// 	_, _, err := net.SplitHostPort(hostport)
+// 	return err
+// }
+
 var rootCfg struct {
 	DebugLog       bool
-	BindPort       int
+	Listen         cmd.ListenAddress
```
GlenDC: I would change this property name to `ListenAddress`

cmd/flag.go
zaibon: compiler complains without the casting

GlenDC: Ah weird, I thought it was about dereferencing, but I tried with some other type aliases in a playground and it doesn't work either.

Weird, I could swear that in situations like this you don't need a cast. Oh well, guess my memory is playing tricks on me here.

PR #373, opened 2017-12-01
add (redesign) client spec

### LINKED ISSUE #367: Client (re)design proposals
After some reflection, and before I start working things more in detail in a spec, I would like to list some (re)design proposal I have in mind:

## Simpler code, More flexibility

Right now we create the client using a policy config struct. It allows you to configure the servers, and the block size. Than it also has some limited configuration for the components. But all in all it is very restrictive.

Now this does seem right more or less for the `zstor` cli client, even though I would still want it to be more flexible even there. However for an API (Golang lib) that is really way to restrictive.

+ We define components, yet some of them we don't use (e.g. we hard-code snappy compress directly);
+ The components have a weird, not really usable design;
+ There is no real pipeline design there, certainly nothing that is really re-used throughout the API;
+ The user cannot define its own components (and certainly cannot use them when using the 0-stor client);
+ The user has little control over the components it does use (e.g. the different compress algorithms have a lot of configuration available, but the user cannot touch any of it);

I propose that we keep the code a lot simpler. I haven't worked out exactly yet, but I certainly would do away with the entire policy thing. I agree that we will define some config for the `zstor` cli client (even though that one will be more flexible than the policy currently is), but for the API this isn't needed.

I would want to have a client:

+ Which does always use a stor-clients for the data servers, I don't think it makes sense to customize this;
+ The IYO config becomes (semi-)optional,as a separate IYO config struct, as the `zstordb` can disable authentication, so this might not be needed;
+ I suppose there will still be some other config as well (e.g. the block size);

However:

+ Rather than creating a metadata client, I think that should be given by the user, and by default it is simply `nil` (and thus we have a more limited API in that case;
+ We should have some generic pipeline object/structure/design, where we can plugin an entire chain of components, with whatever components, in whatever "reasonable" order (it's up to the user to make sure the order make sense, hard to validate this, and why would we care);
+ This way the user can define whatever components she wants, and use whatever components;
+ This also allows the user to create the components objects manually (using the component constructors), making it much easier to expose a per-component configuration for a more fine-grained configuration;
+ None of these components should be hidden behind an interface, even though they will probably implement some component to make them pluggable into the pipeline;

Lastly:

+ As discussed before, we'll want to make the pipeline parallel and streaming (non-blocking), so that we're constantly doing a lot of work at the same time, rather than doing big chunks of work before passing it to the next stage;
+ So that we can do multiple different stores at the same time;
+ The global client interface can just be killed, there is no need for us to expose a Client interface. This object is waay to non-generic for that. If you talk about some kind of small generic object than exposing an interface is fine. But for something the client it doesn't make sense for us. If the user wants to define an interface for it (e.g. to have a stub), than the user can define it herself. We want to expose it simply as a public concrete type directly;

Also the Hasher component is never used. And the replicator never became a component, and instead is just a collection of functions somewhere in the client code.

## Normalize keywords

Currently we use a lot of different keywords (to indicate action), while they all mean the same thing. Here are some examples:

+ Remove == Delete
+ Set == Create == Put == Give

We cannot control the keywords used by APIs that we use (e.g. IYO), but that does not have to mean we expose the same keyword. I would really recommend to choose a keyword for each purpose, and than use that same keyword for all API methods throughout the entire 0-stor codebase. This will make the API simpler to use, as you don't have to worry about whether it was Put or Set or Create.

## Minify Client API

Rename `WriteF` and `ReadF` functions to `Write` and `Read`, and get rid of the wrapper `Write` and `Read functions`.This way the user can write/read whatever, and it's simple enough to define a byte.Buffer if you really want to simply write a byte.Buffer, but I think in most (serious) cases you actually do want to read/write using some kind of streamed type. It also gets rid of the very non-performent `Read` function that is currently defined;

In general you want to keep an API as small as possible. Providing wrappers like this, can easily be done by the user, and should really not be defined by us.

Further I'm still not a big fan of giving the client all these methods though.I think it would be acceptable to have a `IYOClient` that we expose, that provides the permission/namespace functionality. It's still part of the 0-stor client API, but it doesn't have to be all embedded in one structure.

And obviously that wouldn't make developing the `zstor` tool any more difficult, and there it can still look like it's all unified.

GlenDC:
This PR aims to add a specification document,
to put on paper the redesign work to be done in order to turn the
0-stor client into a solid 0-stor component.

Note that the tasks that will flow out of this design, as defined in this spec,
will also involve work on the 0-stor server.

Closes #367


specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+type Data struct {
+    Epoch     int64
+    Key       []byte
+    Chunks    []*ChunkMeta
+    Previous  []byte
+    Next      []byte
+    ConfigPtr []byte
+}
+// ChunkData is new
+type ChunkData struct {
```
zaibon: How do we get to this type ?
I don't see any reference to it

zaibon: How do we get to this type ?
I don't see any reference to it

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+}
+type ObjectStatus uint8 // (enum with String() support)
+type Object struct {
+    key, Data []byte
+    RefList []string
+}
+type ObjectStatusResult struct {
+    Key []byte
+    Status ObjectStatus
+    Error error
```
zaibon: Why do we have an `Error` attribute ? the method already returns errors next to the result.

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+}
+type ObjectResult struct {
+    Object Object
+    Error error
```
zaibon: Same question, why `Error` attribute?

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+}
+
+// Cluster is used in the (main) public `client.Client`,
+// in order to 
+type Cluster interface {
+    // NOTE (QUESTION FOR CHRISTOPHE):
+    // > In the old/current 0-stor client design,
+    // > the given shard does not have to be part of the known shard-list,
+    // > it will simply accept any shard it receives (from a chunk config).
+    // >> IS THIS OK?!?!?!
```
zaibon: Yeah I think it's ok, cause you could have a client that just receive the info from the metadata.
And it needs to be able to get the data out of the shards from the metadata.

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+
+Creating a client would become as simple as calling the following constructor:
+
+```go
+// NewClient creates a new (high-level) 0-stor client.
+// If `blockSize <= 0`, no chunking logic is supported and used (when writing data).
+// This is however not recommended, as it would mean big files have to be
+// loaded completely in memory for processing and storage.
+// 
+// JobCount defines how much (processor) "jobs" or pipelines we run in parallel.
```
zaibon: Not sure about this `jobCount` thing. Cause it doesn't really allow the user to choose exactly how much gorountine will be created, so is there any point at all ?

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+Package: `/client/components/storage`
+
+In it we define the single, distributed and replicated storage logic,
+which is in the current/old design part of the Client itself.
+
+`client/components/storage.Storage`:
+
+```go
+type Storage interface {
+    Write(data ChunkData) (ChunkMeta, error)
```
zaibon: Shouldn't we also need to have a `Status(meta ChunkMeta) (ChunkMeta, error)`. It goes together with `Repair`

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+type Configuration struct {
+    // config defined in our client/itsyouonline package,
+    // and used to create a `client/itsyouonline.Client`.
+    // Required only when the used Data servers require it.
+    IYO itsyouonline.Config `json:"iyo" yaml:"iyo"`
+
+    // Hasher Configuration (used for chunk-key generation),
+    // enabled by default (and cannot be disabled).
+    Hasher struct {
+        // Default: blake2b
```
zaibon: We should not use blak2b as default. We've already shown that the golang implementation is pretty slow.

specs/client.md
```diff
@@ -0,0 +1,724 @@
...
+    // Some Client Functions might not be supported (and return an error),
+    // if this client is not defined though (e.g. Walk).
+    MetaClient meta.Client
+
+    // Optional,
+    // although required if (one of) the used 0-stor servers require it.
+    IYOClient itsyouonline.Client
+
+    // Optional,
+    // Blake2B will be used if not defined.
```
zaibon: We should not use blak2b as default. We've already shown that the golang implementation is pretty slow.

specs/client.md
GlenDC: https://github.com/zero-os/0-stor/blob/ef16faa4363c3515d223ddf79d0dd217b4a69799/specs/client.md#storage

GlenDC: It's to keep all types together, makes more sense than spreading them in cases like this

specs/client.md
GlenDC: The error at the start is if an error happens before we even can create the iterator/channel. While the Error part of the result is because that's the cleanest way (in Golang) to return an error to the user in case the error happens while we're iterating already.

specs/client.md
GlenDC: https://github.com/zero-os/0-stor/pull/373#discussion_r154712553

specs/client.md
GlenDC: Ok, just wanted to make sure.

specs/client.md
GlenDC: It's to give the user the control. It's indeed not a constant (that's hard to do in Golang anyhow, kinda stupid even). See it more as a scalar. Thus if you put a value of 1, that's going to very different than if you put a value of 4. I mean I don't see a big difference between `4` and `3*4`, in reality the user wants 4 jobs, so he gets 4 jobs. We call it jobs for a reason and not goroutines.

Obviously I can remove it from the config, but seems odd to not give the user the choice at least. She might have reasons for example to want to keep the job count as low as possible. And with a job count of 1, we'll probably program things such that there is no parallelism at all even (and in that case everything will be sequential).

specs/client.md
GlenDC: Sure, I don't mind. Which one do you suggest then? `SHA256`?

specs/client.md
GlenDC: No because this storage interface only needs to provide as we need for our own stor client. this is not meant to be a type to be used stand-alone. It's been a couple of days ago, and much has happened in my head since I wrote this document, so I can't recall it all super well. But I'm fairly sure that I didn't need another function in this interface than the one described, in order to  create the internal pipeline model for the client.

specs/client.md
GlenDC: https://github.com/zero-os/0-stor/pull/373#discussion_r154713652

specs/client.md
zaibon: Makse sense to pass error while iterating.

specs/client.md
zaibon: Seems like a reasonable default value yes

specs/client.md
zaibon: OK I understand the idea.
We can keep it, but we should have a default value for it. So it is possible for the user to not have to choose a value.

specs/client.md
GlenDC: Sure thing, for a job-count I usually use https://godoc.org/runtime#NumCPU as the default. I do that for 0-Disk and the 0-stor server as well. This default is used in case `jobCount <= 0`.

specs/client.md
GlenDC: Sure thing, for a job-count I usually use https://godoc.org/runtime#NumCPU as the default. I do that for 0-Disk and the 0-stor server as well. This default is used in case `jobCount <= 0`.

zaibon: ok let's go for these default values

PR #375, opened 2017-12-01
Moves grpc logic

### LINKED ISSUE #327: Move all GRPC code to the server/api/grpc package

chrisvdg:
Closes #327 

State:
api package and the api.API interface are defined.
Logic and tests are moved to new api/grpc package that implements api.API.
Refactored logic and tests and added tests where coverage was missing, There is still some room for improvement though especially for improving the tests.

Also been blocked with TestUpdateReferenceList, removing reference list items doesn't seem to be working (hence the commented out requires at: 
https://github.com/zero-os/0-stor/blob/move-grpc/server/api/grpc/api_test.go#L284
and
https://github.com/zero-os/0-stor/blob/move-grpc/server/api/grpc/api_test.go#L295
)

I think something goes wrong when `SetReferenceList` is called from `RemoveReferenceList` (in server/db/object.go) as the reference list of the `Object`  doesn't change after that call but haven't been able understand exactly why.


GlenDC:
Cool, thank you for your work @chrisvdg.

I'll continue your PR tomorrow. First I need to have PR #376 merged into master, so I can merge master into this PR. Once that's done I'll make sure to finish of your PR.

PR #376, opened 2017-12-02
Redefine db package

### LINKED ISSUE #324: redefine the server/db package and its subpackages
+ [x] define the new DB interface in `/server/db/db.go`;
+ [x] move the `ErrNotFound` error to `/server/db/db.go` as well;
+ [x] redefine the `/server/db/badger/badger.go` file, implementing the new DB interface and following the pointers given as aid in the server spec;
+ [x] redefine the `/server/db/memory/memory.go` file, implementing the new DB interface and following the pointers given as aid in the server spec;
+ [x] adapt the existing `/server/db/db_test.go` unit tests as required, and add missing unit tests as required;
+ [x] add a unit test to ensure that no race conditions are possible
       when updating a value using the `Update` method
       (as to ensure that bugs as described in #264 are no longer possible);
+ [x] ensure package is properly unit tested;
+ [x] ensure unit tests cover at least 80% of the code;

More information can be found at https://github.com/zero-os/0-stor/blob/master/specs/server.md#database.

GlenDC:
Closes #324


GlenDC:
This PR only aims to redesign the DB package as desribed by the server spec. This task was started by @chrisvdg and @katia-e, and I finished it off.

In a future PR we'll make use of the new capabilities provides by the DB. For example we'll make use of the new `db.Update` function to safely update the reference lists for an object. Other improvements will now also be possible (e.g. streaming items (values and/or keys) directly to the GRPC, without having to block until one has been received).

For now however, we only adapted the server code where we had to (e.g. using ListItems instead of List).

PR #384, opened 2017-12-04
Add profiling flags zstor client

### LINKED ISSUE #379: Add profiling flags zstor client
- [ ] profile (CPU)
- [ ] heap
- [ ] block
- [ ] trace

This needs to happen for the current server as the Alpha-8 client to have a comparison.

This feature is part of #296

katia-e:
closes #379

This script will trigger a single profiling mode. In order to trigger all available profiling modes, the script should be called several times. It seems that package `profile` does not support profiling for different modes at the same time, in case several profiling functions are given as arguments of profile.Start(...), only last given profiling mode is triggered.


zaibon:
Yes the title of the PR is a bit confusing. 
I'm not going to merge this and even close it. Cause it's indeed still a WIP and I don't see a point of creating just this right now.

Once the bechmarking client program will be more develop then let's create a new PR for it.

PR #398, opened 2017-12-06
allow empty host in --listen flag of zstordb

### LINKED ISSUE #397: zstordb listen option throws an error if there is no host
```bash
root@64192c87f407:~# zstordb --listen :12345
Error: invalid argument ":12345" for "-L, --listen" flag: host not valid
Usage:
  zstordb [flags]
  zstordb [command]

Available Commands:
  help        Help about any command
  version     Output the version information

Flags:
      --async-write            Enable asynchronous writes in BadgerDB.
      --data-dir string        Directory path used to store the data. (default ".db/data")
  -D, --debug                  Enable debug logging.
  -h, --help                   help for zstordb
  -L, --listen listenAddress   Bind the server to the given host and port. Format has to be host:port, with host optional (default :8080) (default :8080)
      --max-msg-size int       Configure the maximum size of the message GRPC server can receive, in MiB (default 32)
      --meta-dir string        Directory path used to store the meta data. (default ".db/meta")
      --no-auth                Disable JWT authentication.
      --profile-addr string    Enables profiling of this server as an http service.

Use "zstordb [command] --help" for more information about a command.

invalid argument ":12345" for "-L, --listen" flag: host not valid
```

**Version**
```bash
Version: 1.1.0-beta-2
Build: 314b863 2017-12-03T14:23:19+0000
Runtime: go1.9 linux

```

zaibon:
fixes #397

### DIFF
--- a/cmd/flag.go
+++ b/cmd/flag.go
@@ -56,8 +56,10 @@ func (b *ListenAddress) Set(str string) error {
 		return err
 	}
 
-	if ip := net.ParseIP(host); ip == nil {
-		return errors.New("host not valid")
+	if host != "" {
+		if ip := net.ParseIP(host); ip == nil {
+			return errors.New("host not valid")
+		}
 	}
 
 	*b = ListenAddress(str)
--- a/cmd/flag_test.go
+++ b/cmd/flag_test.go
@@ -1,0 +1,63 @@
+package cmd
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestListenAddressFlag(t *testing.T) {
+	tt := []struct {
+		input string
+		err   error
+		value string
+	}{
+		{
+			// default value
+			"",
+			nil,
+			":8080",
+		},
+		{
+			"8080",
+			errors.New("address 8080: missing port in address"),
+			"",
+		},
+		{
+			":8080",
+			nil,
+			":8080",
+		},
+		{
+			"127.0.0.1:8080",
+			nil,
+			"127.0.0.1:8080",
+		},
+		{
+			"badhost:8080",
+			errors.New("host not valid"),
+			"",
+		},
+	}
+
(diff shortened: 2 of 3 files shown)


cmd/flag.go
```diff
@@ -56,7 +56,7 @@ func (b *ListenAddress) Set(str string) error {
 		return err
 	}
 
-	if ip := net.ParseIP(host); ip == nil {
+	if ip := net.ParseIP(host); host != "" && ip == nil {
```
GlenDC: Seem silly to call parseIP even if you're going to ignore the result. Shouldn't this better come:

```go
if host != "" {
   if ip := net.ParseIP(host) // ...
  // ...
}
```

instead?

cmd/flag_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"",
+			nil,
+			":8080",
+		},
+		{
+			":8080",
+			nil,
+			":8080",
+		},
+		{
```
GlenDC: what if I add `8080`? Is that valid? Is that Invalid? Probably should be a test case for as well.

utils/scripts/coverage_test.sh
```diff
@@ -17,3 +17,4 @@ done
 
 test_package_coverage "./client"
 test_package_coverage "./server"
+test_package_coverage "./cmd"
```
GlenDC: By doing this you currently add the entire cmd package and its sub package to coverage, which is not desired, as most of the cmd code is dumb and will never have test, thus resulting in a bad coverage number, even though that's expected.

Can you perhaps change these lines to:

```shell
test_package_coverage "./client/..."
test_package_coverage "./server/..."
test_package_coverage "./cmd"
```

And the change the line

```shell
for d in $(go list "$1/..."); do
```

to

```shell
for d in $(go list "$1"); do
```

please?

zaibon: works for me.

PR #404, opened 2017-12-07
WIP: Adds client benchmarking tool

### LINKED ISSUE #379: Add profiling flags zstor client

### LINKED ISSUE #380: Client benchmark program
As part of #296 we need a program that can benchmark and profile the client with different configurations

* block size
* replication
* distribution
* hashing algo
* encryption algo
* amount of 0-stor servers
* number of (concurrent) clients connected to the servers

chrisvdg:
Adds a 0-stor client benchmarking and profiling tool

Closes #380, closes #379


benchmark/client/benchers/writebencher_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// run limited benchmark
+	wb, err := NewWriteBencher(testID, &sc)
+	require.NoError(err)
+
+	_, err = wb.RunBenchmark()
+	require.NoError(err)
+}
+
+func testGRPCServer(t testing.TB, n int) ([]server.StoreServer, func()) {
```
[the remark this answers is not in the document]
chrisvdg: Yep, from zstor client testing

benchmark/client/main.go
```diff
@@ -0,0 +1,123 @@
...
+	for scID, sc := range clientConf.Scenarios {
+		result := new(benchers.Result)
+		var b benchers.Method
+		var err error
+
+		// define the type of bencher for the method given in scenario
+		benchConstructor, ok := benchers.Methods[sc.BenchConf.Method]
+		if !ok {
+			err = errors.New("benchmark method not found")
+			goto WriteResult
```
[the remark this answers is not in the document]
chrisvdg: If a benchmark scenario fails , the result of that benchmark will have an error field set and continue with the next one.

benchmark/client/benchers/writebencher_test.go
```diff
@@ -0,0 +1,142 @@
...
+		BenchConf: config.BenchmarkConfig{
+			Method:    "write",
+			Duration:  1,
+			KeySize:   5,
+			ValueSize: 25,
+		},
+	}
+
+	// run limited benchmark
+	wb, err := NewWriteBencher(testID, &sc)
```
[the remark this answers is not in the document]
katia-e: by checking timeout before running the next write to 0-stor in `RunBenchmark()`
```
case <-timeout:
	return &wb.result, nil
```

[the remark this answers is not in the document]
chrisvdg: @katia-e , he means that we don't test that a benchmark actually runs for the duration provided to the benchmarker

benchmark/client/benchers/writebencher.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	defer wg.Wait()
+
+	signal := make(chan struct{})
+	defer close(signal)
+
+	go dataAggregator(&wb.result, wb.aggregationInterval, signal, &wg)
```
zaibon: instead I would do
```go
go func(){
dataAggregator()...
wg.Done()
}()
```

benchmark/client/benchers/benchers.go
```diff
@@ -0,0 +1,83 @@
...
+	rand.Seed(time.Now().UnixNano())
+}
+
+const (
+	defaultOperations = 1000000
+)
+
+var (
+	// Methods represent name mapping for benchmarking methods
+	Methods = map[string]func(scenarioID string, conf *config.Scenario) (Method, error){
```
zaibon: Maybe this logic could be hidden into a function. So you don't need this map to be exposed.
Some kind of factory. That's really what you do here.

benchmark/client/benchers/benchers.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Result represents a benchmark result
+type Result struct {
+	Count       int
+	Duration    time.Duration
+	PerInterval []int
+}
+
+func generatedata(len int) []byte {
```
zaibon: rename to `generateData`

benchmark/client/benchers/writebencher.go
```diff
@@ -0,0 +1,119 @@
...
+	if scenario.BenchConf.Operations <= 0 {
+		wb.opsEmpty = true
+		scenario.BenchConf.Operations = defaultOperations
+	}
+
+	// set up data aggregation interval
+	var ok bool
+	wb.aggregationInterval, ok = ResultOptions[scenario.BenchConf.Output]
+	if !ok {
+		wb.aggregationInterval = -1
```
zaibon: What happens in this case ? no aggregation ?

benchmark/client/benchers/writebencher.go
```diff
@@ -0,0 +1,119 @@
...
+
+		// close signal
+		close(signal)
+
+		// wait for data aggregator to return
+		wg.Wait()
+	}()
+
+	start = time.Now()
+	for {
```
zaibon:
```golang
for _, key := range wb.keys {
	select {
	case <-timeout:
		break
	default:
		wb.client.Write(key, wb.value, nil)
		signal <- struct{}{}
	}
}
return &wb.result, nil
```
Like this you don't need the `wb.opsEmpty` at all.
Also I don't know why there infinite for loop was there.

benchmark/client/benchers/benchers.go
```diff
@@ -0,0 +1,83 @@
...
+	data := make([]byte, len)
+	rand.Read(data)
+	return data
+}
+
+//dataAggregator aggregates generated data to provided result
+func dataAggregator(result *Result, interval time.Duration, signal <-chan struct{}) {
```
zaibon: I'm not convinced by this function.

Why do we always do ` totalCount-alreadyCounted`. Why not just have a counter that you reset to 0 after aggregation ?

benchmark/client/benchers/benchers.go
```diff
@@ -0,0 +1,83 @@
...
+	var totalCount int
+	var alreadyCounted int
+
+	defer func(totalCount *int) {
```
zaibon: Why using a pointer here ?

benchmark/client/benchers/benchers.go
chrisvdg: Because we also get the total count as well as the count per interval with this

benchmark/client/benchers/writebencher.go
chrisvdg: if the interval is set < 0 then the interval aggregation is disabled and only the total count will be aggregated

benchmark/client/benchers/writebencher.go
chrisvdg: this will exit then after `defaultOperations` has been reached even if there was no operations provided in the config wich should indicate that the benchmark should olnly exit after the timeout

benchmark/client/benchers/benchers.go
zaibon: yeah I know. but with my solution too. 
I don't see why you need to do `totalCount-alreadyCounted`, which basically gives you the number of request since last time you did an aggregation right ?
Then you do `alreadyCounted = totalCount`.
This logic is so confusing... maybe I didn't understood what you were trying to to. but that shows the complexity of the function.

benchmark/client/benchers/writebencher.go
zaibon: ok I see.

benchmark/client/main.go
```diff
@@ -6,10 +6,10 @@ import (
 	"os"
 	"sync"
 
+	logger "github.com/Sirupsen/logrus"
```
zaibon: import as `log "github.com/sirupsen/logrus"`

benchmark/client/benchers/readbencher.go
```diff
@@ -5,7 +5,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/siddontang/go/log"
+	log "github.com/Sirupsen/logrus"
```
zaibon: import as `log "github.com/sirupsen/logrus"`

benchmark/client/output.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewOutputFormat returns a new OutputFormat
+func NewOutputFormat() OutputFormat {
+	var o OutputFormat
+	o.Scenarios = make(map[string]ScenarioOutputFormat)
+	return o
+}
+
+// OutputFormat represents the output format of a full benchmark
+type OutputFormat struct {
```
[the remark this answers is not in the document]
chrisvdg: They are exported for YAML-marshaling and I guess I could put an anonymous type for the map and make it into one type, but thought it would be cleaner to define a separate type for a scenario result

zaibon:
Will be splitted into multiple PR.

PR #406, opened 2017-12-12
WIP: adds client benchmarking for 1.1.0 alpha 8

chrisvdg:


zaibon:
My bad, We actually need it to work against beta-1 not alpha-8

PR #408, opened 2017-12-13
add tags to enable profiling

zaibon:


cmd/zerostorserver/main.go
```diff
@@ -108,6 +113,17 @@ func main() {
...
+		cli.StringFlag{
+			Name:        "profile-kind",
+			Usage:       "Select the type of profiler to enable. Can be 'cpu', 'mem','trace','block'.",
+			Destination: &profileKind,
+		},
+		cli.StringFlag{
+			Name:        "profile-ouput",
+			Usage:       "Output directory where the profile result will be written. Required --profile-kind to be set.",
+			Destination: &profileOutput,
+			Value:       ".",
```
chrisvdg: What does the file name look like using this default?
Can't it be `zstorprofile.pprof` or something?

zaibon: it depends on the mode of the profile enable.
cpu.pprof, mem.pprof. block.pprof, etc...

PR #410, opened 2017-12-13
add flags to enable profiling on zstordb

### LINKED ISSUE #378: Add profiling flags zstor server
- [ ] profile (CPU)
- [ ] heap
- [ ] block
- [ ] trace

This needs to happen for the current server as the Alpha-8 server to have a comparison.

This feature is part of #296

zaibon:
fixes  #378


cmd/zstordb/commands/root.go
```diff
@@ -78,6 +79,34 @@ func rootFunc(*cobra.Command, []string) error {
...
 	}
 
+	if rootCfg.ProfileMode != "" {
+		stat, err := os.Stat(rootCfg.ProfileOutput)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				log.Fatalln(err)
+			}
+			if err := os.MkdirAll(rootCfg.ProfileOutput, 0660); err != nil {
+				log.Fatalf("fail to create profile output directory: %v\n", err)
```
GlenDC: this should be returned as an error, no need to panic in CMD code, error results are supported

cmd/zstordb/commands/root.go
```diff
@@ -78,6 +79,34 @@ func rootFunc(*cobra.Command, []string) error {
...
+			}
+		}
+		if !stat.IsDir() {
+			log.Fatalf("profile-output (%s) is not a directory\n", rootCfg.ProfileOutput)
```
GlenDC: this should be returned as an error, no need to panic in CMD code, error results are supported

cmd/zstordb/commands/root.go
```diff
@@ -78,6 +79,34 @@ func rootFunc(*cobra.Command, []string) error {
...
+		}
+
+		switch rootCfg.ProfileMode {
+		case "cpu":
+			defer profile.Start(profile.NoShutdownHook, profile.ProfilePath(rootCfg.ProfileOutput), profile.CPUProfile).Stop()
```
GlenDC: can you please keep your lines short, these lines are pretty long, making it  a bit difficult to read in narrow windows

cmd/zstordb/commands/root.go
```diff
@@ -103,6 +132,8 @@ var rootCfg struct {
 	DebugLog       bool
 	ListenAddress  cmd.ListenAddress
 	ProfileAddress string
+	ProfileMode    string
```
GlenDC: Instead of working with a raw string, can we perhaps define an enum-flag for this in the flags package? so we can also use it for the other command line tools, and so we don't have to compare strings all the time? Especially in the inline-error-prone way as is done now.

cmd/zstordb/commands/root.go
```diff
@@ -78,6 +79,34 @@ func rootFunc(*cobra.Command, []string) error {
...
+		case "cpu":
+			defer profile.Start(profile.NoShutdownHook, profile.ProfilePath(rootCfg.ProfileOutput), profile.CPUProfile).Stop()
+		case "mem":
+			defer profile.Start(profile.NoShutdownHook, profile.ProfilePath(rootCfg.ProfileOutput), profile.MemProfile).Stop()
+		case "trace":
+			defer profile.Start(profile.NoShutdownHook, profile.ProfilePath(rootCfg.ProfileOutput), profile.TraceProfile).Stop()
+		case "block":
+			defer profile.Start(profile.NoShutdownHook, profile.ProfilePath(rootCfg.ProfileOutput), profile.BlockProfile).Stop()
+		default:
+			log.Fatalf("profile-kind %s not supported\n", rootCfg.ProfileMode)
```
GlenDC: this should be returned as an error, no need to call a fatal in CMD code, error results are supported

cmd/zstordb/commands/root.go
```diff
@@ -78,6 +79,34 @@ func rootFunc(*cobra.Command, []string) error {
 		}()
 	}
 
+	if rootCfg.ProfileMode != "" {
+		stat, err := os.Stat(rootCfg.ProfileOutput)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				log.Fatalln(err)
```
GlenDC: this should be returned as an error, no need to panic in CMD code, error results are supported

cmd/flag.go
```diff
@@ -70,3 +71,40 @@ func (b *ListenAddress) Set(str string) error {
 func (b *ListenAddress) Type() string {
 	return "listenAddress"
 }
+
+//ProfileMode is a string representing a profiling mode
+//which can be used as a flag for Cobra command
+type ProfileMode string
```
GlenDC: While this work, I was thinking to go for something like:

```go
type ProfileMode int8

const (
  ProfileModeCPU ProfileMode = iota
  ProfileModeMem
  ProfileModeBlock
  ProfileModeTrace
)

func (p *ProfileMode) String() string {
   // convert int8 -> string (using "" for invalid ProfileMode)
}

func (p *ProfileMode) Set(str string) error {
   // convert string -> int8
}

func (p *ProfileMode) Type() string {
  return "profileMode"
}
```

What do you think? It's a small detail, but I think it's cleaner than having to do so many strings checks all the time.

Perhaps I'm focussing too much on this detail so, as this code is really not that performance-critical, so I don't think we'll ever feel this performance-loss.  But still.

Either way, I just add it as a comment, but will still approve it. As the rest looks good. So feel free to go for my proposal, or if you're happy with what you have, you can simply ignore this comment.

cmd/zstordb/commands/root.go
```diff
@@ -129,6 +172,10 @@ func init() {
 		&rootCfg.DB.Dirs.Meta, "meta-dir", ".db/meta", "Directory path used to store the meta data.")
 	rootCmd.Flags().StringVar(
 		&rootCfg.ProfileAddress, "profile-addr", "", "Enables profiling of this server as an http service.")
+	rootCmd.Flags().VarP(
+		&rootCfg.ProfileMode, "profile-mode", "", "Enable profiling mode, one of [cpu, mem, trace, block]")
```
GlenDC: would be good to have these options defined (as a slice?) as part of the flag definition, so this never gets out of sync, and so we can simply reuse it here in the description (and wherever else we use that flag)

zaibon:
@GlenDC can you validate the last commit of this PR please. 

I noticed a possible dead lock situation. I think it's fixed, but maybe you have another approach to solve it.

GlenDC:
> I noticed a possible dead lock situation. I think it's fixed, but maybe you have another approach to solve it.

Yeah I see how it can deadlock though, I don't agree with your fix though, let me amend a commit, and I'm also going to improve some other code of this PR, if you don't mind.

PR #411, opened 2017-12-14
adds zstor setup lib for bench orchestrator

chrisvdg:


benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
+import subprocess
+import tempfile
+import shutil
+from random import randint
+
+# SetupZstor is responsible for maanaging a zstor setup
+class SetupZstor:
+    def __init__(self, zstors=2, etcds=1, no_auth=True, jobs=0, start_zstor_port=1200, start_etcd_port=1300):
```
zaibon: instead of passing all the arguments in the constructor. I would instead move it in the run methods.
By itself, the SetupZstor object doesn't need anything to be created.

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+        self.etcds = etcds
+        self.no_auth = no_auth
+        self.jobs = jobs
+        self.start_zstor_port = start_zstor_port
+        self.start_etcd_port = start_etcd_port
+        self.zstor_nodes = []
+        self.etcd_nodes = []
+
+    # run both etcd and zstor servers
+    def run(self):
```
zaibon: move the argument of the constructor in this method

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+        self.run_etcd_servers()
+        self.run_zstordb_servers()
+
+    # stop both etcd and zstor servers
+    def stop(self):
+        self.stop_zstordb_servers()
+        self.stop_etcd_servers()
+
+    # start zstordb servers
+    def run_zstordb_servers(self):
```
zaibon: add required argument for the zstordb

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+        for i in range(0, self.zstors):
+            db_dir = tempfile.mkdtemp()
```
zaibon: This needs to be configurable. add argument for it with default None. If user doesn't specify it you can create a temp directory like you do now.

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+
+            if self.no_auth:
+                args.append("--no-auth")
+            
+            self.zstor_nodes.append(subprocess.Popen(args, stderr=subprocess.PIPE))
+
+    # stop zstordb servers
+    def stop_zstordb_servers(self):
+        for node in self.zstor_nodes:
+            node.terminate()
```
zaibon: here you need to also call `node.wait()`
Otherwise you'll end up with zombie processes. `node.termintate()` only send the signal that ask the process to stop.
The parent process still need to wait for the child process to finish to properly clean it up.

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+        self.zstor_nodes = []
+
+    # start etcd servers
+    def run_etcd_servers(self):
```
zaibon: same as for zstordb, move argument to this method

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,93 @@
...
+                "--initial-cluster", init_cluster,
+                "--data-dir", db_dir,
+            ]
+
+            self.etcd_nodes.append(subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE))
+
+    # stop etcd servers
+    def stop_etcd_servers(self):
+        for node in self.etcd_nodes:
+            node.terminate()
```
zaibon: same remark. need to wait for the child to finish

benchmark/orchestrator/lib/zstor_local_setup.py
zaibon: Also can you add some argument to enable the profiling mode. cpu,mem,block,trace

PR #412, opened 2017-12-15
Client daemon

### LINKED ISSUE #68: client daemon

iwanbk:
Fixes #68


zaibon:
Need to be reworked to map to the new changes server side

PR #414, opened 2017-12-17
Adds benchmarking client lib

chrisvdg:


benchmark/orchestrator/lib/bench_client.py
```diff
@@ -0,0 +1,22 @@
+import lib.zstor_local_setup as zstor
+import subprocess
+
+class Bench_client:
+    def __init__(self, ):
+        pass
+    
+    def run(self, exe="./client", profile=None, profile_dir="profile_client", config="client_config.yaml", out="bench_result.yaml"):
```
zaibon: should not received an `exe` argument but use the name of the bechmark client directly `zbench`

PR #416, opened 2017-12-18
ensure label for JWT-enabled client is correct

GlenDC:
The (zstordb) server (JWT Auth Middleware) requires the label to be in the format of `<org>_0stor_<namespace>`. When we run a server that doesn't use the JWT verification, it doesn't care, so the label can simply be `<namespace`>

Having that in mind, I have put some validation in the JWT code, to ensure that the label is in this format, as we need to be able to take just the `<namespace>` part at that point, as the IYO code in our repo, relies on the fact that we pass in just that part to `CreateJWT`, otherwise the token won't be created.

Having said that, perhaps it makes sense to enforce this label format in that case, in the entire code, config and all. Regardless if the IYO verification is configured and used, or not. Because right now, having just this PR (as it is), we get this invalid-label error, only when trying to apply a method on a (zstordb) shard, which is kinda late.

So having this error earlier might be user-friendly. On top of that, it's also kinda weird that the label can be in any format when JWT-verification is disabled, while otherwise it is OK.

Anyway, I can amend something to this PR should it be needed, or we can leave it as it is, up to you @zaibon.

On top of that, I also request the admin permission, should it be available:

+ this is required for the GetNamespace datastor method;
+ it also seems to make it possible to allow for examle a write method, even though you're not in the write group, as long as you're in the admin group;


zaibon:
The user should not really have to know about this namespace format. At least from a config point of view. It should only specify the organization and the namespace. And the code should create the correct format out of it.

In the case authentification is disabled, well then I don't really know how we could enforce that since we don't have to specify an organization.

Regarding the admin scope. It should indeed be required, cause if you're admin, you have all rights, all the time.

GlenDC:
> The user should not really have to know about this namespace format. At least from a config point of view. It should only specify the organization and the namespace. And the code should create the correct format out of it.

Yeah you're right. I just was thinking about this as well.

I'm going to change this PR a bit. I think it's fine to just have the namespace in the config, we'll just have to make sure to change the label we pass to the ztordb server, when we're using JWT verification.

You just wrote what I had in mind as well, what a coincidence :)

PR #417, opened 2017-12-18
don't make codecov fails if commit has a decrease in coverage

zaibon:
don't make codecov fails if commit has a decrease in coverage but overall project has more then 75% of coverage


zaibon:
@GlenDC I think this config should make it. If you want the detail, see https://docs.codecov.io/docs/commit-status

I don't know why codecov think I changed something in `client/metastor/encoding/proto/metadata.pb.go` or `server/api/grpc/schema/zstor.pb.go` though

GlenDC:
> I don't know why codecov think I changed something in client/metastor/encoding/proto/metadata.pb.go or server/api/grpc/schema/zstor.pb.go though

I wouldn't worry about that, unless you find out why. I noticed too in other PR's, that sometimes codecov thinks that a coverage for a given file changed. Not sure why it does that. Maybe its relative to something, dunno, who knows how it calculates those percentages.

.codecov.yml
```diff
@@ -1,4 +1,9 @@
 coverage:
+  status:
+  project:
```
GlenDC: is invalid, you forgot an indention for all this stuff, you would notice if you use their validation HTTP service:

```
curl --data-binary @.codecov.yml https://codecov.io/validate
Invalid value None (NoneType): must be bool or must be Mapping (at coverage['status'])
```

PR #425, opened 2017-12-23
generate zstor object key on server side

### LINKED ISSUE #225: Text file with repeated content failed to be downloaded

### LINKED ISSUE #421: provide better gen-key-templates for datastor objects created by 0-stor client
When writing a "file" using the https://github.com/zero-os/0-stor/blob/7ee85f40ec92cf1feca8896d696b27520cb9528d/client/client.go#L129-L131 function, we do so giving a key. This is however the key of the file/metadata, and not the keys of the single or multiple datastor objects dat will be used for storage of the actual data of that file.

The keys for the single or multiple datastor objects, containing the data of that "file", are automatically generated, using a cryptographic hasher. This is done by checksumming the raw data (as in the data before any processing is applied) and on top of that, when a private key is provided, a signature is created using that checksum and the private key as input.

While this is great on paper, it brings perhaps, an unforeseen consequence with it (as I hope this is just due to oversight).

It means that when storing multiple files, using the same configuration, blocks will be shared, if they are stored on the same (zstordb) shard. Meaning several things:

+ When deleting a "file" using the 0-stor client, it simply deletes all datastor objects stored for that file, no matter if that object is also used by other "files" (this is a big problem, and is very likely to hit users right in the face very soon);
+ Reference lists will simply be overwritten, if a datastor object gets created on a shard, where  that content already existed, previously (this is only the case if the second file also makes use of the reference list, otherwise the reference list will not be touched, which would mean that this second file suddenly has a reference list (which it will retrieve when reading the file), even though it didn't make use of references at all, which is a related, but different, problem);

For the Reference lists I already started an issue, #420, as it fits in a bigger picture, of rethinking the reference-support, as there are more problems with it, than just the one thing mentioned here.

So for this issue I would like to focus mostly on the first problem, that is, that a file might suddenly lose data, because it (unknowingly by the user?) shares objects between files.

Still, we might assume that the user owns the shards they use, and also chooses the private key, but the problem would still be there. The user might have many files, and as long as you have enough files, which are large enough, and a small enough object size (e.g. 4KiB), you'll very likely start sharing objects, very soon.

The simplest, most naive, solution would be to prefix all datastor objects, when generated via the 0-stor master client, with a namespace + "file"-key. It can however never be as simple as this, as this would mean that having deduped files, is no longer possible, should this be something the 0-stor user wants to have. Still, all is possible, by extending the configuration, but I don't think the solution can be as simple as this, as for sure there would be other issues coming with that approach.

First of all I think it needs to be clear what kind of options/use-cases we want to open up for our users. Do we ever want to support deduped content? if so, this needs to be taken into account, just like all other requirements that we establish. Once that's clear, we should try to come up with a robust solution, that can handle all this, without creating problems as we have with the current design.

GlenDC:
Fixes #421 (in 8da04fc06a7731a2abf490da5f559355f1b9ffa0):

+ adds new vendored deps: hashicorp/golang-lru (to implement badger sequence cache);
+ update badger db vendor dep (master, as to get sequence feature);
+ adds SetScoped method to server DB interface (used to create an object without pre-defining the key);
+ test SetScoped method in server DB code;
+ update GRPC zstordb interface to make the object key generated,
      using db's SetScoped under the hood;
+ integrate and update code;
+ hash from hasher is now stored as an extra piece of data (used as extra validation)
      (it can also be used by user-defined features to implement a deduped feature);
+ add any missing unit tests and cleanup code;

This feature also fixes #225, as it turned out to be an issue in the fact that we were storing all the different erasure-coded parts using the same key, and thus when you have multiple chunks with the same (hash-defined) key, you end up overwriting parts that shouldn't be overwritten. In 816e94e061ec6e09e205b84eea9d00bcf3d9e45e unit tests have been added, some of which using the repeated-content file as linked in issue #225.

All stored objects are now unique at all times, should deduplication be desired, than this has  to be implemented on a layer from above. How that could be done is yet to be defined.

Lastly this PR also removes update function from server db interface: the update function is no longer required by the zstordb server and was only introduced in this milestone, thus better clean this up before this becomes a legacy contract.


server/db/memory/memory_test.go
```diff
@@ -71,6 +70,226 @@ func TestSetGet(t *testing.T) {
...
+	for output := range outputCh {
+		require.True(t, strings.HasPrefix(output.Key, output.ScopeKey))
+
+		m, ok := allOutput[output.ScopeKey]
+		require.True(t, ok)
+
+		_, ok = m[output.Key]
+		require.False(t, ok)
+
+		// TODO: UN-COMMENT
```
GlenDC: need to make a separate issue for this, might be a bug in badger that we need to help investigate

PR #448, opened 2018-01-05
add daemon implementation

### LINKED ISSUE #68: client daemon

GlenDC:
fixes #68:

+ add daemon which can be started using the 'zstor daemon' command;
+ add daemon GRPC interface with 4 interfaces:
   NamespaceService, FileService, DataService, MetadataService;
+ add unit tests for all the services, locally as well as entire round-trip cycles;


daemon/api/grpc/data_api.go
```diff
@@ -0,0 +1,353 @@
...
+		return nil, rpctypes.ErrGRPCNoLocalFS
+	}
+
+	filePath := req.GetFilePath()
+	if len(filePath) == 0 {
+		return nil, rpctypes.ErrGRPCNilFilePath
+	}
+
+	// open the file
+	file, err := openFileToRead(filePath)
```
zaibon: Why do we need this `openFileToRead` method ? It just wraps `return os.Open(path)` I don't see the benefit

GlenDC: It's for unit tests, so we don't need actual I/O operations. In the unit tests we really want to isolate our operations as much as possible, so that's why.

PR #453, opened 2018-01-08
strip out all non-std proto code from daemon.proto

GlenDC:
this should allow people to generate a daemon client
for other languages as well (e.g. using the official tool provided by google)


GlenDC:
@zaibon AFAIK the code as it is, is fine.

At the moment it is only possible to have a nil-pointer panic, should you try to marshal a GRPC message (struct) which has a nil-pointer property. However that is only possible if there's a bug in the server/client (wrapper) code, as those control the marshaling completely, not the end-user. So that's fine. During unmarshaling all is handled already by the generated GRPC code.

So can you please either approve, or provide me with a review.

PR #465, opened 2018-01-11
Benchmark orchestrator beta2

katia-e:


katia-e:
@GlenDC , We finished changes to adjust benchmark for beta-2.

benchmark/benchers/readbencher.go
```diff
@@ -0,0 +1,132 @@
...
+		interval = time.Second
+	}
+
+	var (
+		tick            = time.Tick(interval)
+		start           time.Time
+		counter         int64
+		intervalCounter int64
+		result          = &Result{}
+		maxIteration    = len(rb.keys)
```
GlenDC: change to `len(rb.keys) - 1`, so you don't have to do it each iteration (the increment)

benchmark/benchers/readbencher.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+
+	start = time.Now()
+	for i := 0; ; i++ {
```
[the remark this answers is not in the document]
GlenDC: change `for i := 0; ; i++ {` to `for i := 0; i == maxIteration; i++ {`

not need for `>=` by the way, as you are doing increments by one, so there is no reason that it should ever miss the `maxIteration`...

benchmark/benchers/readbencher.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		if timeout == nil && i >= maxIteration-1 {
+			break
+		}
+	}
+	result.Duration = Duration{time.Since(start)}
+	result.Count = counter
+	if intervalCounter != 0 {
+		result.PerInterval = append(result.PerInterval, intervalCounter)
```
GlenDC: why are you ignoring `0`? Seems like an interesting case to know about as well.

benchmark/benchers/utils_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/zero-os/0-stor/client/pipeline"
+	"github.com/zero-os/0-stor/client/pipeline/processing"
+	"github.com/zero-os/0-stor/server/api"
+	"github.com/zero-os/0-stor/server/api/grpc"
+	"github.com/zero-os/0-stor/server/db/badger"
+)
+
+// newEmbeddedMetaServer creates new embedded metadata (etcd) server
+func newEmbeddedMetaServer() (*embeddedMetaServer, error) {
+	tmpDir, err := ioutil.TempDir("", "etcd")
```
GlenDC: please don't use `ETCD` for unit tests. Use the in-memory metastor DB please :(

check other unit tests in this codebase for inspiration

benchmark/benchers/readbencher_test.go
```diff
@@ -0,0 +1,87 @@
+package benchers
+
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
GlenDC: please divide imports in 3 groups, where each group is separated with a newline. Obviously it's possible that you might only have 1 or 2 groups.


```
import (
   <Golang STD>

   <0-stor>

    <other>
)
```

this is also done in the other parts of the codebase, and it's a tiny detail, but it really makes it much clearer, without any effort.

benchmark/benchers/utils_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	cfg := embed.NewConfig()
+	cfg.Dir = tmpDir
+
+	capnslog.SetGlobalLogLevel(capnslog.CRITICAL)
+	// listen client URL
+	// we use tmpDir as unix address because it is a simple
+	// yet valid way to generate random string
+	lcurl, err := url.Parse("unix://" + filepath.Base(tmpDir))
```
GlenDC: also don't use unix sockets in unit tests, will break on Operating systems like windows.

benchmark/benchers/writebencher.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		log.Errorf("Error creating client: %v", err)
+		return nil, fmt.Errorf("Failed creating client: %v", err)
+	}
+
+	return wb, nil
+}
+
+//RunBenchmark implements Method.RunBenchmark
+func (wb *WriteBencher) RunBenchmark() (*Result, error) {
```
GlenDC: a lot of the same comments apply as the ones I gave for `(*ReadBencher).RunBenchmark`, check comments there and apply them here as well, where applicable

benchmark/benchers/writebencher.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// set up data aggregation interval
+	interval, ok := ResultOptions[wb.scenario.BenchConf.Output]
+	if !ok {
+		interval = time.Second
+	}
+
+	var (
+		tick            = time.Tick(interval * 1)
```
GlenDC: why are you multiplying by one? What is the point?

benchmark/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+// Validate validates a scenario
+func (sc *Scenario) Validate() error {
+	return sc.BenchConf.Validate()
+}
+
+// SetupClientConfig sets up the client.Client for a benchmark.
+// Removes IYO fields. (Benchmarks uses no-auth zstordb's)
+// Sets random namespace if empty
+func SetupClientConfig(c *client.Config) {
+	// empty IYO fields
```
GlenDC: I don't agree with this at all. I think it should be up to the user to define whether or not the servers are `--no-auth` or not. Thus if the user defines the `IYO Config`, authentication is used, otherwise not.

Because I think it to be very useful to be able to run the benchmarks with the authentication active, as it can impact the results a lot, which is info I need to know as a user! As that is what is how its used  in production. So if the IYO system is slow (and/or we have a bad caching strategy, or none at all), the results might be a lot worse. So there is no point in hiding this, unless the users asks to hide this.

benchmark/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	c.IYO = itsyouonline.Config{}
+
+	// set namespace if not provided
+	if c.Namespace == "" {
```
GlenDC: That's a bit weird, I don't know why we do this. But OK, I can live with it for now.

benchmark/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+		c.Namespace = "b-" + randomSuffix(4)
+	}
+}
+
+func randomSuffix(n int) string {
+	var s string
```
GlenDC: why don't you use `rand.Read` here?

benchmark/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	Output     string `yaml:"result_output"`
+	Duration   int    `yaml:"duration"`
+	Operations int    `yaml:"operations"`
+	Clients    int    `yaml:"clients"`
+	KeySize    int    `yaml:"key_size" validate:"nonzero"`
+	ValueSize  int    `yaml:"value_size" validate:"nonzero"`
+}
+
+// Validate validates a BenchmarkConfig
+func (bc *BenchmarkConfig) Validate() error {
```
GlenDC: I really want these validators to be private and part of the deserialization.

GlenDC: As in practice they'll always be created from files, so this is really not a problem.

benchmark/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	if bc.Duration <= 0 && bc.Operations <= 0 {
+		return fmt.Errorf("duration or operations was not provided")
+	}
+
+	return validator.Validate(bc)
+}
+
+// FromReader returns client config from a given reader
+func FromReader(r io.Reader) (*ClientConf, error) {
```
GlenDC: Why from Reader? I'm fairly sure we only ever read it from a file, so I think you can keep it as simple as that.

benchmark/config/testconfigs/invalidDurOpsConf.yaml
```diff
@@ -0,0 +1,32 @@
+scenarios:
```
GlenDC: please move these yaml config files used for tests under a subdir within fixtures

cmd/zstorbench/README.md
```diff
@@ -0,0 +1,223 @@
...
+Benchmark client provides tools for benchmarking and profiling zstor client for various scenarios.
+
+Configuration for benchmarking scenarios should be given in YAML format (see [example](#yaml-config-file) of the config file bellow).
+Benchmarking program outputs results for all provided scenarios to a single output file in YAML format (see [example](#yaml-output-file) of the output file bellow). 
+
+
+## Getting started
+
+In order to start benchmarking program at least one [zstor server](https://github.com/zero-os/0-stor/blob/master/docs/gettingstarted.md) have to be set up.
+When running a zstor server `--no-auth` flag has to be given for the sake of performance testing. This option allows to skip the authentification step. Client config still has to provide all necessary fields in order to pass the validation and successfully create a new zstor client.
```
GlenDC: This should be removed, it should really be up to the user whether or not this is true, as your performance is impacted by the authentication, hopefully but not too much, but still, needs to be visible. If the user however specifies no IYO config, than it's fine.

cmd/zstorbench/README.md
```diff
@@ -0,0 +1,223 @@
...
+```
+zstordb --no-auth -D --listen 127.0.0.1:2379:12345 --data-dir stor1 --meta-dir stor1
```
GlenDC: Your listen address is wrong here, copy paste mistake?

You are also using the same dir for data and meta, which is not something you should do AFAIK, at least use 2 subdirs

cmd/zstorbench/README.md
```diff
@@ -0,0 +1,223 @@
...
+  -h, --help                   help for this command
+      --out-benchmark string   path and filename where benchmarking results are written (default "benchmark.yaml")
+      --out-profile string     path where profiling files are written (default "profile")
+      --profile-mode string    enable profiling mode, one of [cpu, mem, trace, block]
+
+```
+
+Start benchmarking with default input/output files
+``` 
+./client
```
GlenDC: you should really give the benchmark tool a better name, as client is very bad of a name

benchmark/benchers/benchers.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	// Methods represent name mapping for benchmarking methods
+	benchers = map[string]BencherCtor{
+		"read":  NewReadBencher,
+		"write": NewWriteBencher,
+	}
+	// ResultOptions represent name mapping for benchmarking methods
+	ResultOptions = map[string]time.Duration{
+		"per_second": time.Second,
```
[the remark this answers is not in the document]
chrisvdg: this is for the result.Perinterval slice, each entry in the slice represents the counted operations in that 1 second interval, milliseconds seems a bit short for that no?

benchmark/benchers/benchers.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func init() {
+	// seed random generator
+	rand.Seed(time.Now().UnixNano())
+}
+
+const (
+	// defaultOperations is set when BenchConf.Operations was not provided
+	defaultOperations = 1000
```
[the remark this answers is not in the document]
chrisvdg: Playing around, mostly so it wouldn't take too long to write data to the zstordb for rsetting up the reading benchmark.
Suggestions for this number are welcome ofc

benchmark/benchers/benchers.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var (
+	// Methods represent name mapping for benchmarking methods
+	benchers = map[string]BencherCtor{
+		"read":  NewReadBencher,
+		"write": NewWriteBencher,
```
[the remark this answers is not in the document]
chrisvdg: We didn't implement it because read and write were the most important operations to be benchmarked, there was also talk of maybe removing the delete operation when designing the zstorbench in future

benchmark/benchers/readbencher.go
```diff
@@ -0,0 +1,132 @@
...
+// RunBenchmark runs the read benchmarker
+func (rb *ReadBencher) RunBenchmark() (*Result, error) {
+	if rb.client == nil {
+		log.Error("zstor client is nil when trying to run a read bencher")
+		return nil, fmt.Errorf("zstor client is nil")
+	}
+
+	var timeout <-chan time.Time
+	if rb.scenario.BenchConf.Duration <= 0 {
+		timeout = nil
```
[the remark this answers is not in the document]
chrisvdg: if a 0 or negative number is provided no timeout will be set for the benchmark, this is valid as the benchmark can also be ended after a certain number of operations

benchmark/benchers/benchers.go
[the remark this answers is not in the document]
GlenDC: Perhaps, but as far as I could see, the user could choose that no? I mean it's not like you will look accuracy by going for seconds, it will just mean you have more decimals I suppose.

I guess you can leave it in for now, I'm not too strict on this. Let's see the results, and we can always modify it later on.

benchmark/benchers/benchers.go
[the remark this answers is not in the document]
GlenDC: I think it's fine, make sure it's documented in the function/constructor comments that this default is used though, as well as in the config godocs and externals docs. If not already.

benchmark/benchers/benchers.go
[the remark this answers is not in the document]
GlenDC: Sure, but it's a small effort to measure deletion. I think it's still a useful thing to measure. But I suppose we can add it in the future if we really want it.

benchmark/benchers/readbencher.go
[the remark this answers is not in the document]
GlenDC: Sure, I can accept that, except that you do break the loop, after 1 iteration always, when your timeout variable is nice. Due to your `timeout == nil` check :) But if you apply my other suggestions, that will no longer be the case and  thus this condition is fine.
Even so, you can just keep this code as:

```go
var timeout <-chan time.Time
if rb.scenario.BenchConf.Duration > 0 {
	 timeout = time.After(time.Duration(rb.scenario.BenchConf.Duration) * time.Second)
}
```

As in Golang, it will use sane zero/nil values where possible. And thus `timeout` will be nil as soon as you declare it (`var timeout <-chan time.Time`).

benchmark/benchers/readbencher.go
```diff
@@ -0,0 +1,132 @@
...
+	} else {
+		timeout = time.After(time.Duration(rb.scenario.BenchConf.Duration) * time.Second)
+	}
+
+	// set up data aggregation interval
+	interval, ok := ResultOptions[rb.scenario.BenchConf.Output]
+	if !ok {
+		interval = time.Second
```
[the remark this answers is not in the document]
chrisvdg: Changing this to where `Output` is invalid, it would not keep track of the counts per interval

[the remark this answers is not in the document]
GlenDC: What?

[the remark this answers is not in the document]
chrisvdg: removing the ok check, so that if the `BenchConf.Output` is not found/invalid, `interval` will be 0 (and `tick` `nil`) and therefor will not keep track of the operations per interval,

benchmark/benchers/utils_test.go
chrisvdg: Any pointers where I can find them? Can't seem to find usable inspiration for using the memory metastor

benchmark.yaml
```diff
@@ -0,0 +1,49 @@
+scenarios:
```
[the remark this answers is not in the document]
GlenDC: Any comment on this?

benchmark/README.md
```diff
@@ -0,0 +1,109 @@
...
+To start the benchmarking provide a [config file](#orchestrator-config-file) for `benchmark orchestrator` (`0-stor/benchmark/orchestrator/`) and run using optional parameters:
+``` bash
+python3 orchestrator --conf orchCong.yaml --out report
+```
+
+Here are the options of the orchestrator:
+``` bash
+optional arguments:
+  -h, --help            help for orchestrator
+  -C  --conf string     path to the config file (default rchConfig.yaml)
```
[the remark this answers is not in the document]
GlenDC: Any comment on this?

benchmark/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Orchestrator config file
+Config file for the `orchestrator` consists of two parts:
+
+  * `template` represents the template of the config file for the benchmark client.
+
+  * `benchmarks` contains set of the benchmark parameter.
+  Multiple benchmarks can be added to the final report. 
+
+The config for each benchmark is marked by the `prime_parameter` and inside there can be an optional `second_parameter` defined.
```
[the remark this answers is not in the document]
GlenDC: Any comment on this?

benchmark/benchers/utils_test.go
GlenDC: Sure, you'll have to create a 0-stor client manually I'm afraid, copying some of the code, but I'll show you how it's done in a bit, need to do some other things first.

benchmark/config/config.go
GlenDC: Any comments on this?

benchmark/config/testconfigs/invalidDurOpsConf.yaml
GlenDC: Any comments on this?

cmd/zstorbench/README.md
katia-e: This readme is outdated, I'll change it now

benchmark.yaml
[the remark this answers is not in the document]
chrisvdg: It was added by accident indeed

benchmark/config/config.go
chrisvdg: It was suggested to do so in an earlier code review as that would make it more flexible, so the config input doens't necessarily needs to come from a file

benchmark/config/config.go
chrisvdg: I don't see how you can do this more simply with rand.Read

benchmark/config/testconfigs/invalidDurOpsConf.yaml
chrisvdg: fixtures as in?

chrisvdg: nvm, got it!

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,247 @@
...
+            # define empty benchmark
+            self.prime = Benchmark()
+            self.second = Benchmark()                                               
+
+def port2int(port):
+    try:
+        port = int(port)
+    except:
+        raise ValueError("wrong port format")        
+    return port
```
muhamadazmy: missing new line

benchmark/orchestrator/orchestrator.py
```diff
@@ -0,0 +1,120 @@
...
+                    report.add_timeplot()
+            
+            # add results of the benchmarking to the report
+            report.add_aggregation()
+            config.restore_template()
+    except StopIteration:           
+        print("Benchmarking is done")
+
+if __name__ == '__main__':
+    main(sys.argv[1:])
```
muhamadazmy: missing new line at end of the file

katia-e:
The code is ready for final review

katia-e:
library for `packet.net` deployment is removed from this brunch to a separate brunch. Now only packages for local tests are included. PR for benchmarking `beta2` on  `packet.net` will be opened ones this PR is merged.  We will not push any more commits in this PR until it's requested.

benchmark/benchers/writebencher.go
```diff
@@ -0,0 +1,115 @@
...
+	"time"
+
+	"github.com/zero-os/0-stor/benchmark/config"
+	"github.com/zero-os/0-stor/client"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+//WriteBencher represents a writing benchmarker
+type WriteBencher struct {
```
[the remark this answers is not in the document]
chrisvdg: Couldn't really come up with something to do it right, though thinking it could be done with a callback kind of thing?

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,266 @@
+"""
```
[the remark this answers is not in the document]
katia-e: @GlenDC , I agree on that, I will try to restructure and simplify it, but if you point out what is done wrong at least from design perspective, it will make it faster for me.

benchmark/assets/example_report/report.md
```diff
@@ -0,0 +1,51 @@
...
+      distribution:
+        data_shards: 3
+        parity_shards: 1
+      encryption:
+        private_key: ab345678901234567890123456789012
+        type: aes
+
+```
+![Fig: throughput vs parameter](fig1.png)
+ <h3> Throughput, byte/s: </h3>
```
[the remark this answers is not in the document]
katia-e: I tried markdown tables and got impression that they give less control, so I did html. Perhaps for what we do, markdown is enough, so I'll change to it.

benchmark/orchestrator/lib/config.py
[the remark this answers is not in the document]
GlenDC: Sure, but let's not block the PR for now on that. If you want we can do this as a future PR, with me pointing it out when I'm visiting the office in Lochristi, might be easier, and also gives us a change to meet one another.

benchmark/README.md
```diff
@@ -27,13 +27,13 @@ optional arguments:
...
   
-The orchestrator sets up a benchmarking environment (zstordb and etcd metadata servers) and uses [`zstorbench`](../cmd/zstorbench/README.md) to collect the performance data. The orchestrator will generate the config files with scenarios and environment configuration for [`zstorbench`](../cmd/zstorbench/README.md) and process the output (of multiple zstorbench runs) to generate a report.
+The orchestrator sets up a benchmarking environment (zstordb and etcd metadata servers) and uses [`zstorbench`](/cmd/zstorbench/README.md) to collect the performance data. The orchestrator will generate the config files with scenarios and environment configuration for [`zstorbench`](/cmd/zstorbench/README.md) and process the output (of multiple zstorbench runs) to generate a report.
 
 ### Orchestrator config file
 Config file for the `benchmark orchestrator` consists of three parts (main fields):
 
   * `benchmarks`: configuration of what to benchmark
-  * `template`: template configuration for [`zstorbench`](../cmd/zstorbench/README.md) client
+  * `template`: template configuration for [`zstorbench`](cmd/zstorbench/README.md) client
```
GlenDC: You forgot the initial forward slash...

benchmark/README.md
```diff
@@ -43,7 +43,7 @@ The config for each benchmark is marked by the `prime_parameter` field and an op
 If only `prime_parameter` is given, the `orchestrator` creates a plot of the throughput versus values in `range` of the `prime parameter`.
 If both `prime_parameter` and `second_parameter` are given, a few plots will be combined in the output figure, one for each value in `range` of `second_parameter`.
 
-The `template` field represents the template of the config file for [`zstorbench`](../cmd/zstorbench/README.md).
+The `template` field represents the template of the config file for [`zstorbench`](cmd/zstorbench/README.md).
```
GlenDC: You forgot the initial forward slash...

Make sure to test, you can easily do on your machine. As I think you're using visual code, you can do so using a preview window (CTRL+K+V, I think it is on Linux).

cmd/zstorbench/README.md
```diff
@@ -14,12 +14,12 @@ make install
 In order to start a benchmarking run, all necessary [zstor servers](https://github.com/zero-os/0-stor/blob/master/docs/gettingstarted.md) have to be set up.
 
 Start the benchmarking
-``` bash
+```
```
GlenDC: no this can be bash though, was just referring to your options code-field, just here below

cmd/zstorbench/README.md
```diff
@@ -29,12 +29,12 @@ zstorbench -C config.yaml --out-benchmark benchmark.yaml
 
 
 Start benchmarking with optional input/output files
-``` bash
+```
```
GlenDC: can be bash as well

cmd/zstorbench/README.md
```diff
@@ -29,12 +29,12 @@ zstorbench -C config.yaml --out-benchmark benchmark.yaml
...
 zstorbench --conf "input_config.yaml" --out-benchmark "output_benchmark.yaml"
 ```
 
 Start benchmarking and profiling
-``` bash
+```
```
GlenDC: can be bash as well

cmd/zstorbench/config.yml
```diff
@@ -1,13 +1,14 @@
...
-      iyo:  # if empty or omitted, the zstordb servers set up for the benchmark 
+      iyo:  # If empty or omitted, the zstordb servers set up for the benchmark 
             # need to be run with the no-auth flag.
-            # For benching with authentication, provide it with valid itsyou.online credential
+            # For benching with authentication, provide it with valid itsyou.online credentials
         organization: "bench_org"
         app_id: "an_iyo_bench_app_id"
         app_secret: "an_iyo_bench_app_secret"
-      namespace: adisk
+      namespace: adisk # If IYO credentials are provided, this needs to be a valid IYO namespace,
```
GlenDC: and existing?

benchmark/bencher/bencher.go
```diff
@@ -0,0 +1,210 @@
...
+	"math"
+	"math/rand"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/zero-os/0-stor/benchmark/config"
+	"github.com/zero-os/0-stor/client"
+)
+
+func init() {
```
GlenDC: You really made it much more complicated than needed. This was just supposed to be an internal change. The API user doesn't have to know code is shared. You can still have separate Read/Write constructors. Now you have a lot of interfaces (by the way only expose interfaces in Golang in very specific cases, otherwise don't do it, it's your homework to figure the details of this comment), having this entire new config as well just makes things more complex. Really it's not needed, and in the end it doesn't give you anything extra, it just makes things more complex.

cmd/zstorbench/cmd/root.go
```diff
@@ -1,114 +1,107 @@
...
-	RootCmd.Flags().StringVar(&BenchmarkFlags.profileMode, "profile-mode", "", "enable profiling mode, one of [cpu, mem, trace, block]")
+	RootCmd.PersistentFlags().BoolVarP(&rootCfg.debugLog, "debug", "D", false, "Enable debug logging.")
+	RootCmd.Flags().StringVarP(&rootCfg.confFile, "conf", "C", "zstorbench_config.yaml", "path to a config file")
+	RootCmd.Flags().StringVar(&rootCfg.benchmarkOutPath, "out-benchmark", "benchmark.yaml", "path and filename where benchmarking results are written")
+	RootCmd.Flags().StringVar(&rootCfg.profileOutPath, "out-profile", "./profile", "path where profiling files are written")
+	RootCmd.Flags().VarP(&rootCfg.profileMode, "profile-mode", "", rootCfg.profileMode.Description())
 }
 
-func root(cmd *cobra.Command) {
+func root(cCmd *cobra.Command) {
```
GlenDC: cCmd is really a bad name. And anyhow I don't think you're ever accessing it. In Golang you can just keep your parameters anonymous if you don't use them, as in `func root(*cobra.Command) {`. `root`, on itself, is by the way also a really bad name.

GlenDC: That function should by the way also get the `func rootFunc(*cobra.Command, []string) error` signature, and be used directly as the `RunE` property of that root cmd, because what you're doing now is a bit weird/bad.

Issue #466, opened 2018-01-12
create organization if it doesn't exist yet for zstor create namespace command


GlenDC:
Currently we assume that the given organization already exists, during the execution of the `zstor create namespace` command. However it could be useful to be able to provide the automated feature to create the organization should it not exist yet, this way, the user doesn't have to go to the website first to create the organization.

Perhaps this shouldn't be the default behavior, and instead be behind some `-f/--force` flag, not sure about this one. But I do know it is a feature that is pretty useful. Or perhaps we could put it behind a `y(es)/n(o)` question, and provide a `-y/--yes` flag, to answer all such questions as yes, should they arise.

We should probably also log an info-level log, describing we're creating the organization programmatically, if we do so, because it is missing.

GlenDC:
@zaibon would be happy to hear your opinion about this one

zaibon:
I think it's a good idea. If we can provide a full 0-stor experience without the need for the user to go to the IYO webstie at all it's a good idea.
CLient would still have to get an app ID and secret, but from there he can manage everything from the zstor CLI.

I don't have a problem of making this the default behavior even. Do we know if IYO allow this and what kind of rights it needs to create an organization ?

GlenDC:
I think IYO allows it, there is definitely a function for available in the generated IYO client. I don't think any special permissions are needed.

The only problem is however that the userID is not known, as we use appID/appSecret, while this is required as part of the POST request (see `POST /organizations` in https://itsyou.online/apidocumentation).

Somehow we'll need to get to know the userID of the user. Getting that userID using the API isn't possible either i think, so it would be required somehow as a config param.

On top of that organizations can also be owned by an organization instead of a user I think, supporting that is however optional I suppose.

GlenDC:
The error status code returned by `CreateNewSubOrganization (`POST /organization/{globalid}`) is a vague `403` (`Forbidden`) though, so doesn't really say it's because it's because the organization itself doesn't exist.

Even so, I guess in our code we can try to interpret it as a organization-not-exists error, create that organization and see if that fixes it. If however that does not fix it, than I suppose we should return the original 403 error (and cleanup the mess we made on the way?).

lafont-e:
[...]

PR #473, opened 2018-01-17
Create organization when needed (issue #466)

lafont-e:
CreateNamespace() now checks that Organization and 0stor organizations are available, if not, then they are created.

Two commits 
commit 8879d43e555a7bbdcaca1ae0ab2859d8f35ae30f
commit 8eb2b562e8376b918180707f3a3facda2221f45a
commit cf3734559775e5174678e4ca2add7441a086691

the 1st one is a esthetic change as it only changes the format of an error message I did not like

### DIFF
--- a/client/itsyouonline/itsyouonline.go
+++ b/client/itsyouonline/itsyouonline.go
@@ -120,7 +120,7 @@ func createSubOrganization(c *Client, org, suborg string) error {
 
 	if err != nil {
 		if resp.StatusCode == 409 {
-			return fmt.Errorf("[Error] namespace already exists code=%v, err=%v", resp.StatusCode, err)
+			return fmt.Errorf("[Error] %s exists code=%v, err=%v", body, resp.StatusCode, err)
 		}
 		return fmt.Errorf("code=%v, err=%v", resp.StatusCode, err)
 	}
--- a/client/itsyouonline/itsyouonline.go
+++ b/client/itsyouonline/itsyouonline.go
@@ -115,6 +115,7 @@ func (c *Client) CreateJWT(namespace string, perm Permission) (string, error) {
 func createSubOrganization(c *Client, org, suborg string) error {
 	body := org + "." + suborg
 	sub := itsyouonline.Organization{Globalid: body}
+	fmt.Println("Creating ", body)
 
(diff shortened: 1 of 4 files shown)


cmd/zstor/commands/file.go
```diff
@@ -31,6 +31,25 @@ import (
 var fileCmd = &cobra.Command{
 	Use:   "file",
 	Short: "Upload or download files to/from (a) 0-stor server(s).",
+	// overwrite the namespace config property, if it is given as a flag by the user
```
GlenDC: all stuff in your diff from the bottom, until here, is already stuff that's on master. So I think you forgot to rebase your master prior to working on this branch on your own master.

Also you really should learn to work on a feature branch (e.g. `git checkout -b feat-466`) rather than on master, as this is just messy, and also prevents you from working on multiple features at once, which will be expected from you soon enough. On top of that I want you to push this branch directly to github.com/zero-os/0-stor rather than to your own fork, as this makes it possible for others (e.g. me) to collaborate with you on the same branch. For projects like this, I expect you to use your fork only for WIP/throw-way branches, or in case you work from multiple work environments, such that you can easily sync WIP between machines.

client/itsyouonline/itsyouonline.go
```diff
@@ -122,52 +140,43 @@ func (c *Client) CreateNamespace(namespace string) error {
...
-	if err != nil {
-		if resp.StatusCode == http.StatusConflict {
-			// provide a more user-friendly error message for known/expected errors
-			return fmt.Errorf("namespace %[1]s (%[2]s.%[1]s) already exists", namespace, namespaceID)
+
+	// Create c.cfg.Organization if non existent
+	org := c.cfg.Organization
+	if resp.StatusCode == 403 { // Forbiden, organization does not exist mainly
+		organization := itsyouonline.Organization{Globalid: org}
+		_, resp, err := c.iyoClient.Organizations.CreateNewOrganization(organization, nil, nil)
```
GlenDC: Are you sure you confirmed this worked? As I could swear you also need to include a list of members and owners? I'll test it tonight myself to confirm. If it works like this, in the correct way, all the better I suppose.

PR #476, opened 2018-01-18
Adds benchmark orchestration scripts

chrisvdg:
Before this can be merged:

- [x] #475 needs to be merged into master and this branch rebased (adding zstorbench)
- [x] Implement code review from beta2 that can be applied here as well


chrisvdg:
I've added a shebang to the orchestrator script, as far as I know, those always always had to be on the first line, is there a way around this with the copyright header script?

GlenDC:
> I've added a shebang to the orchestrator script, as far as I know, those always always had to be on the first line, is there a way around this with the copyright header script?

Yeah i think we can apply the following patch to the master branch:

```diff
diff --git a/utils/scripts/copyright_check.sh b/utils/scripts/copyright_check.sh
index 72497fe..5b21359 100755
--- a/utils/scripts/copyright_check.sh
+++ b/utils/scripts/copyright_check.sh
@@ -6,7 +6,7 @@ PYCRFILE=$(dirname "$0")/copyright.py.header
 
 FAILED_FILES=""
 
-# python license check
+# golang license check
 for f in $(find "$DIR" -name "*.go" | grep -v vendor | grep -v pb.go | grep -v pb_test.go); do
     diff <(sed -n -e '1,15p' "$f") <(sed -n -e '1,15p' "$GOCRFILE")
     if [ $? -ne 0 ]; then
@@ -15,8 +15,12 @@ for f in $(find "$DIR" -name "*.go" | grep -v vendor | grep -v pb.go | grep -v p
 done
 
 # python licence check
-for f in $(find "$DIR" -name "*.py" | grep -v vendor | grep -v generated| grep -v test_suite|grep -v benchmarker); do
-    diff <(sed -n -e '1,13p' "$f") <(sed -n -e '1,15p' "$PYCRFILE")
+for f in $(find "$DIR" -name "*.py" | grep -v vendor | grep -v generated| grep -v test_suite|grep -v utils); do
+    SOURCE_LINES="1,13p"
+    if [ "$(head -n 1 $f)" = "#!/usr/bin/python3" ]; then 
+        SOURCE_LINES="3,15p" 
+    fi;
+    diff <(sed -n -e $SOURCE_LINES "$f") <(sed -n -e '1,13p' "$PYCRFILE")
     if [ $? -ne 0 ]; then
         FAILED_FILES="$FAILED_FILES\n  > $f"
     fi
```

That should allow your shebang thing to be first, and have it being separated with a white line with the copyright header.

GlenDC:
Will review this once PR #465 is finished and merged into master. And until all the (required) changes in there are also applied/ported over to here.

GlenDC:
Master now allows your python hash bangs @chrisvdg and have restarted the failed tests where that fixes your issues. So now at least your checks are all green :)

benchmark/README.md
```diff
@@ -0,0 +1,155 @@
+# zstor benchmark
+
+`zstor benchmark` provides tools to analyse the performance of `zstor`.
+
+## Getting started
+
+Download the `zstor`repository.
```
zaibon: missing a space here

benchmark/README.md
```diff
@@ -0,0 +1,155 @@
...
+```bash
+git clone https://github.com/zero-os/0-stor.git
+
+# or 
+go get github.com/zero-os/0-stor
```
zaibon: you can use `go get -d` to only download

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+                            raise InvalidBenchmarkConfig("for '{}' cannot convert val = {} to type {}".format(key,val,parameter_type))
+                        return True
+            return False
+        if not replace(self.template, key_id, val):
+            raise InvalidBenchmarkConfig("parameter %s is not supported"%key_id)
+
+    def restore_template(self):
+        """ Restore initial zstor config """
+
+        self.template = deepcopy(self._template0)
```
zaibon: not sure deepcopy is needed here. dict are kept by reference anyway, so just doing `self.template = self._template0` should do the trick

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+        # write scenarios to a yaml file
+        with open(file_name, 'w+') as outfile:
+            yaml.dump(output, outfile, default_flow_style=False, default_style='')
+
+    def update_deployment_config(self):
+        """ 
+        Fetch current zstor server deployment config
+                ***specific for beta2***
+        """
+        try:
```
zaibon: 2 things I don't like here.
- except that catch all, you don't really know what you catch, it's bad practice
- the distribution is mandatory. Which mean you can't run test with replication or even without any replication or distribution.

I would do something like
```python
self.datastor = self.template['zstor_config']['datastor']
if 'distribution' in self.datastor['pipeline']:
    distribution = self.datastor['pipeline']['distribution']
    self.data_shards_nr = int(distribution['data_shards']) + int(distribution['parity_shards'])
elif 'replication' in self.datastor['pipeline']:
    self.data_shards_nr = int(self.datastor['pipeline']['replication'])
else:
    # if nothing specified, just use 1 ztrordb
    self.data_shards_nr = 1
```

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def __init__(self, parameter={}):
+       
+        if parameter:
+            self.id = parameter.get('id', None)
+            self.range = parameter.get('range', None)
+            if not self.id or not self.range:
+                raise InvalidBenchmarkConfig("parameter id or range is missing")
+            
+            if isinstance(self.id, dict):
```
zaibon: This is a bit cryptic, a comment explaining what's going would be nice

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+                            return True
+                    return False
+                if not contain(PARAMETERS_DICT, self.id):
+                    raise InvalidBenchmarkConfig("parameter {0} is not supported".format(self.id))
+            else:
+                if self.id not in PARAMETERS:
+                    raise InvalidBenchmarkConfig("parameter {0} is not supported".format(self.id))
+            
+            try:
+                self.range = split("\W+", self.range.strip(',-!?.'))
```
zaibon: no need for the split function from the re package, split is a method on type string already.
```python
self.range.strip(',-!?.').split(',')
```

But that raise the question, why is range a string and not a list in the config in the first place ?
If you want to support both notation you can do something like
```python
if isinstance(self.range, str):
    self.range = self.range.split(',')
....
```

benchmark/orchestrator/lib/report.py
```diff
@@ -0,0 +1,346 @@
...
+                outfile.write("\n![Fig: throughput vs parameter]({0})".format(fig_name))
+
+        # add the table of the data sets
+        self._add_table()
+
+    @staticmethod
+    def humanize_bitrate(value_in_bytes):
+        ''' Shortens large values of bitrate to KB/s or MB/s '''
+
+        kbyte, mbyte, byte = 'KB/s', 'MB/s', 'Byte/s'
```
zaibon: 1KB = 1000 bytes
1KiB = 1024 bytes
Same for MB and MiB

benchmark/orchestrator/lib/zstor_local_setup.py
```diff
@@ -0,0 +1,204 @@
...
+            self.cleanup_dirs.extend((db_dir, md_dir))
+
+            port = str(pick_free_port())
+            self.data_shards.append('%s:%s'%(Base,port))
+
+            args = ["zstordb",
+                    "--listen", ":" + port,
+                    "--data-dir", db_dir,
+                    "--meta-dir", md_dir,
+                    "--jobs", str(jobs),
```
zaibon: what happens if jobs is 0?

benchmark/README.md
GlenDC: Indeed, on top of that I would also add the `-u` flag, so it updates the package(s), in case they were already present on the local FS.

chrisvdg: fair enough but it still throws that error for me using those flags

zaibon: then maybe you should not document this method at all.
Or try to install `github.com/zero-os/0-stor/cmd/zstorbench` instead

benchmark/orchestrator/lib/config.py
katia-e: If I try it without `deepcopy`  `self._template0` changes together with `self.template `, I wand them to be different dictionaries

zaibon: ok make sense

benchmark/orchestrator/lib/config.py
katia-e: If I except the catch, how do I handle i `zstor_config` or `datastor` are not in the dictionary? Or it's alright is I just let it fail?

zaibon: > if zstor_config or datastor are not in the dictionary? 

This raise a KeyError, so yo can catch that.
The problem here is that you masque the real error. so it's harder to know what was wrong.
catching specific excpetion type allow you to controle and report more detail errors

katia-e: @zaibon , what is `replication` ?

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+    def wait_local_servers_to_start(self):
+        """ Check whether ztror and etcd servers are listening on the ports """
+
+        addrs = self.deploy.data_shards + self.deploy.meta_shards
+        servers = 0
+        timeout = time.time() + 20
+        while servers < len(addrs):
+            servers = 0
+            for addr in addrs:
+                port = ':%s'%split(':', addr)[-1]
```
GlenDC: Isn't the new and accepted way of formatting string in python now `{0}".format(...)`?

benchmark/orchestrator/lib/config.py
GlenDC: Agree, some comments would be helpful

benchmark/orchestrator/lib/config.py
GlenDC: Range should already be a (YAML) list, having it be a string first isn't helpful in any way, already discussed in the mini meeting after the SCRUM this morning.

benchmark/orchestrator/lib/config.py
GlenDC: Replication means we replicate the same data on multiple shards. Meaning we write the same piece of data on x amount of nodes. When reading that data we only need to read from one of those written shards, not all of them.

This used to be a separate setting, but is now implicitly defined by only defining the `data_shards` and leave the `parity_shards` property either undefined or less then 1.

benchmark/orchestrator/lib/config.py
```diff
@@ -0,0 +1,302 @@
...
+            self.data_shards_nr=distribution['data_shards'] + distribution['parity_shards']
+        except:
+            raise InvalidBenchmarkConfig("distribution config is not correct")
+        try:
+            self.metastor  = self.template['zstor_config']['metastor']
+            self.meta_shards_nr = self.metastor['meta_shards_nr']
+        except:
+            raise InvalidBenchmarkConfig("number of metastor servers is not given")
+        
+        self.no_auth = True
```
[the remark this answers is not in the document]
zaibon: This made me tick also, but you can't do that cause here we still want to get IYO info in a variable.
this would put it inside `self.no_auth`, which doesn't sounds rgiht.
But it can be refactor to be `self.iyo_token`, and test the no auth if `self.iyo_token` is none or not

benchmark/orchestrator/lib/zstor_local_setup.py
GlenDC: As documented, if `--jobs` (JobCount) is `0` or negative, the default JobCount will be used, as defined by the zstor client Golang API.

benchmark/orchestrator/lib/config.py
zaibon: both are correct.
I tend to use .format when you have multiple time the same variable to replace.
using `%` is a bit faster then .format although here speed this is not important

GlenDC:
@katia-e and/or @chrisvdg, let me (and @zaibon) know when this PR is ready for a final review.

benchmark/README.md
```diff
@@ -0,0 +1,155 @@
...
+    datastor:
+      pipeline:
+        block_size: 4096
+        compression:
+          mode: default
+        distribution:
+          data_shards: 2
+          parity_shards: 1
+    metastor:
+      meta_shards_nr: 2
```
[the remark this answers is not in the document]
chrisvdg: Since the orchestrator sets up the zstor environment (locally or on packet) it doesn't make sense to have the ip addresses in the orchestrator's  zstor config as you can't know them in advance

benchmark/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Make sure `etcd` is installed if benchmarking with `etcd` as metadata server. On how to install `etcd`, check the [etcd documentation](https://coreos.com/etcd/docs/latest/).
+Zstor requires	etcd 3.2.4 or any higher stable release.
+
+Install python dependencies:
+```bash
+pip3 install -r benchmark/orchestrator/requirements.txt
+
+# matplotlib dependency 
+sudo apt-get install python3-tk
```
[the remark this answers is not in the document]
chrisvdg: added for dnf, yum and homebrew, but couldn't test it as I don't have a macOs machine, so could you verify those steps?

benchmark/assets/example_orchestrator_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+- prime_parameter:    # (required) primary parameter of the benchmark *
+    id: value_size    # (required) id of the primary parameter that is being benchmarked
+    range: [1024000, 2048000, 4096000] # (required) values of the primary parameter
+  second_parameter:    # (optional) secondary parameter of the benchmark *
+    id: key_size       # (required)id of the secondary parameter that is being benchmarked
+    range: [48000, 96000]  # (required) values of the primary parameter 
+template: # config for benchmark client (zstorbench)
+  zstor:  # zstor configuration
+    datastor:   # (optional, will use 1 data shard with default values if omitted) datastor config.
+      pipeline:
```
GlenDC: is optional as well, or at least it should be if it isn't yet

katia-e: @GlenDC , yes now `pipeline` is optional, as well as `datastor`

GlenDC: Yes, ok, I tested it out and it does seem to work indeed without. Reading the code confirms why this works.

Issue #468, opened 2018-01-12
Deleting namespaces


elafont:
* I have a User1 who is the owner of test260.0stor organization (all tests done with zstor)
* I create a new namespace eraseme
```
$ zstor namespace create eraseme  
INFO[0000] namespace "eraseme" created        
```
* User1 has all the rights on eraseme
```
$ zstor namespace permission get User1  eraseme
Read: true
Write: true
Delete: true
Admin: true
```
* When trying to delete the namespace eraseme, or any other created 
```
$ zstor namespace delete  eraseme  
INFO[0000] namespace "eraseme" deleted 
```

* All seems right, but going to the IYO Web site shows that *eraseme* is still live and working 
```
$ zstor file upload aFile  -n eraseme 
INFO[0000] Setting namespace config to [eraseme]        
INFO[0000] data from "aFile" uploaded as key = "aFile"
```

No matter how many times I ask for a delete, the namespace is never erased.

The config files used are the sames as in #467.

lafont-e:
Small code to test, only uses itsyouonline client
```go
// test code for  itsyouonline issue 690: deleting a namespace

package main

import (
	"fmt"
	iyo "github.com/zero-os/0-stor/client/itsyouonline"
	"log"
	"os"
)

const (
	user1        = "lafonte@greenitglobe.com"
	user1Id      = "otzMc8WXFwBqdfU_YiPiMQBgTkeU"
	user1Secret  = "S_ZiZNGWZOe_7ZzkTdbkmCV_ZiZp"
	organization = "test260"
)

var (
	iyo1 = iyo.Config{
		Organization:      organization,
		ApplicationID:     user1Id,
		ApplicationSecret: user1Secret}
)

func usage() {
	log.Fatalf("Usage: %s [create|delete] <namespace> ", os.Args[0])
}

func createNs(iyoCl *iyo.Client, ns string) {
	fmt.Println("Creating namespace ", ns)
	err := iyoCl.CreateNamespace(ns)
	if err != nil {
		log.Fatalf("Error creation of namespace %q failed: (%s)", ns, err)
	}

	fmt.Printf("namespace %q created\n", ns)
	fmt.Println()
}

func removeNs(iyoCl *iyo.Client, ns string) {
	fmt.Println("Removing namespace ", ns)
	err := iyoCl.DeleteNamespace(ns)
	if err != nil {
		log.Fatalf("Error removing of namespace %q failed: (%s)", ns, err)
	}

[...]

PR #481, opened 2018-01-19
Itsyouonline vendor update

zaibon:
iyo had an update to fix issue #468 and this code applies the
changes needed to the 0-stor code tree


GlenDC:
OK, now I understand why you closed the original PR of @lafont-e, @zaibon :)

PR #497, opened 2018-01-31
Adds proposal for benchmark configs

chrisvdg:
Proposal of benchmark configs for beta2
Results of the proposal ran locally can be found here:
https://github.com/chrisvdg/0-stor/tree/1.1.0-beta-2-benchmarked/benchmark/benchmarks_local
https://github.com/chrisvdg/0-stor/tree/1.1.0-beta-2-benchmarked/benchmark/profiling_local


GlenDC:
Cool, will play with it today, and review it when I did.

chrisvdg:
Keep in mind that with changes from #498 the range values of the benchmarks need encapsulated with `[ ]` to turn them into lists

benchmark/benchmarks_local/datashards_vs_metashards/read_bench.yaml
```diff
@@ -0,0 +1,28 @@
+benchmarks:
```
zaibon: Not sure this is usefull.
Cause even if you change the data_shards number, the distribution configuration remain the same.
So you will always hit the same number of zstordb

benchmark/benchmarks_local/datashards_vs_metashards/write_bench.yaml
```diff
@@ -0,0 +1,28 @@
+benchmarks:
```
zaibon: Not sure this is usefull.
Cause even if you change the data_shards number, the distribution configuration remain the same.
So you will always hit the same number of zstordb

benchmark/benchmarks_local/read_vs_write/bench.yaml
```diff
@@ -0,0 +1,34 @@
+benchmarks:
+- prime_parameter:
```
zaibon: value size is not really something we care about here.
So i would jus keep the second test

benchmark/profiling_local/compression/gzip_bench.yaml
```diff
@@ -0,0 +1,33 @@
+# Benchmarks snappy compression types
```
zaibon: snappy =>  gzip

benchmark/profiling_local/compression/lz4_bench.yaml
```diff
@@ -0,0 +1,33 @@
+# Benchmarks snappy compression types
```
zaibon: snappy => lz4

benchmark/benchmarks_local/datashards_vs_metashards/read_bench.yaml
GlenDC: I guess that's a failure in the bench config to not be able to handle this in sync?

GlenDC: Nevermind, it seems the orchestrator python (config) code already adapts the number of zstordb servers based on the data+parity shards (if you define (any of) those). So I think just the data+parity shards test is good enough. We can also test replication tests in range, by just defining the data shards. But the config as it is now, isn't useful at all. Will adapt the configs.

benchmark/benchmarks_local/blocksize_vs_valsize/read_bench.yaml
```diff
@@ -0,0 +1,28 @@
+benchmarks:
+- prime_parameter:
+    id: block_size   
+    range: [1048576, 2097152, 4194304, 1048576]
```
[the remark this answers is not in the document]
GlenDC: This can be resolved as part of #507.

Issue #494, opened 2018-01-30
Error marshalling CompressionMode


chrisvdg:
Similar issue to #493

When generating the output for zstorbench with compression disabled, it always throws the following error:
`Error: '' is not a valid CompressionMode value`

This error is throw from the  `MarshalText`method from `processing.CompressionMode` as it returns an error when the string representation of a CompressionMode is an empty string, the string representation of a disabled compression mode is also an empty string.

`MarshalText` is called when marshaling the client config (that includes this processing config) to a YAML file as the benchmarks include the client and benchmark configurations in the results.

chrisvdg:
suggesting the following changes

```diff
diff --git a/client/processing/compression_mode.go b/client/processing/compression_mode.go
index 7a66d84..505d9e7 100644
--- a/client/processing/compression_mode.go
+++ b/client/processing/compression_mode.go
@@ -63,20 +63,21 @@ func (cm CompressionMode) String() string {

 var (
        _CompressionModeValueToStringMapping = map[CompressionMode]string{
-               CompressionModeDisabled:        "",
-               CompressionModeDefault:         _CompressionModeStrings[:7],
-               CompressionModeBestSpeed:       _CompressionModeStrings[7:17],
-               CompressionModeBestCompression: _CompressionModeStrings[17:],
+               CompressionModeDisabled:        _CompressionModeStrings[:8],
+               CompressionModeDefault:         _CompressionModeStrings[8:15],
+               CompressionModeBestSpeed:       _CompressionModeStrings[15:25],
+               CompressionModeBestCompression: _CompressionModeStrings[25:],
        }
        _CompressionModeStringToValueMapping = map[string]CompressionMode{
                "": CompressionModeDisabled,
-               _CompressionModeStrings[:7]:   CompressionModeDefault,
-               _CompressionModeStrings[7:17]: CompressionModeBestSpeed,
-               _CompressionModeStrings[17:]:  CompressionModeBestCompression,
+               _CompressionModeStrings[:8]:    CompressionModeDisabled,
+               _CompressionModeStrings[8:15]:  CompressionModeDefault,
+               _CompressionModeStrings[15:25]: CompressionModeBestSpeed,
+               _CompressionModeStrings[25:]:   CompressionModeBestCompression,
        }
 )

-const _CompressionModeStrings = "defaultbest_speedbest_compression"
+const _CompressionModeStrings = "disableddefaultbest_speedbest_compression"

```

PR #498, opened 2018-01-31
Benchmark beta2 small fixes

katia-e:
Couple small fixes have been dome after review in #476. Among those:

- format of range in the benchmark orchestrator config is changed to list (was string)
- improved formating of the output plot (location of bar labels)


katia-e:
### Note
Current change is setting default config for `beta-2` orchestration in the same way as in `master` version  #476. Now only minimal config needed to start benchmarking. However, if `compression mode` is not given, the benchmark will fail because of the issue similar to #494 that was fixed for master only.

GlenDC:
> However, if compression mode is not given, the benchmark will fail because of the issue similar to #494 that was fixed for master only.

@katia-e can you open a PR for beta-2 branch, to merge those changes that fixed it in master?

chrisvdg:
@katia-e, the fix is in this commit: https://github.com/zero-os/0-stor/commit/4815d1474e9a590093ec601f83fa900905d1dc5f

PR #509, opened 2018-02-13
List namespaces

### LINKED ISSUE #505: add list of namespaces to itsyouonline.Client
So user knows all namespaces he has

iwanbk:
Fixes #505


client/itsyouonline/itsyouonline.go
```diff
@@ -146,6 +146,39 @@ var (
...
+		if resp.StatusCode == http.StatusForbidden {
+			return nil, ErrForbidden
+		}
+
+		return nil, err
+	}
+
+	var namespaces []string
+	for _, child := range org.Children {
+		if child.Globalid != orgName {
```
GlenDC: Can this really happen? I would think it only ever returns maximum one child with the globalID being the orgName, I don't see what else it would return. Either way, I guess this check doesn't hurt.

client/itsyouonline/itsyouonline.go
```diff
@@ -146,6 +146,39 @@ var (
...
+			continue
+		}
+		for _, c := range child.Children {
+			elems := strings.Split(c.Globalid, ".")
```
GlenDC: You still miss validation to ensure that the first 2 elements are indeed `<org>` and `0stor`.

client/itsyouonline/itsyouonline.go
```diff
@@ -146,6 +146,39 @@ var (
...
+			if len(elems) == 3 {
+				namespaces = append(namespaces, elems[2])
+			}
+		}
+	}
+
+	return namespaces, nil
```
GlenDC: An error `NotFound` should be returned in case no namespaces are found for the given client and org config.

zaibon: should it ?
Having no namespace to list is not really an error

GlenDC: That's debatable, either way if it's a known error you can easily filter it out should you not want it as an error. But OK, I'm not anal about this one, if you want to keep it as a non-error I can live with that.

iwanbk:
> I'm also surprised not to see this new method exposed to the NamespaceService of the daemon? Is this not needed than?

i was completely forgot about the daemon. fixed it in latest commit

client/itsyouonline/itsyouonline.go
iwanbk: fixed

cmd/zstor/commands/namespace.go
```diff
@@ -79,6 +79,31 @@ func getNamespacePosArg(maxArgs int, args []string) string {
...
+		if err != nil {
+			return err
+		}
+
+		namespaces, err := iyoCl.ListNamespaces()
+		if err != nil {
+			return fmt.Errorf("list namespaces failed: %v", err)
+		}
+
+		log.Infof("namespaces:")
```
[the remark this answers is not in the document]
iwanbk: fixed

Issue #500, opened 2018-02-01
size of actual data in the metadata


iwanbk:
To be used as minio backend, we need to know the size of actual data in the metadata.

From my test, both meta.Size and chunk.Size is the size of processed (compressed & encrypted) data.

As far as i can see, it is a mandatory feature @zaibon

iwanbk:
> As far as i can see, it is a mandatory feature @zaibon

When downloading data, the client do these:
- get size of the data -> this is why we need this feature
- download the data using `content-length` header with value = size of the data. I guess it will use `offset` and `length` if the size is too big, as described at #499

GlenDC:
Can you link me to the minio docs that describe these mandatory features you're talking about?

> From my test, both meta.Size and chunk.Size is the size of processed (compressed & encrypted) data.

That is correct, however, if you would like, we could implement the actual _total_ data in the metadata, it shouldn't be that hard.

iwanbk:
See `GetObjectInfo`  of https://godoc.org/github.com/minio/minio/cmd#ObjectLayer.
It return https://godoc.org/github.com/minio/minio/cmd#ObjectInfo which has `Size` field.

Which is needed when downloading object as described at https://github.com/zero-os/0-stor/issues/500#issuecomment-362213123

iwanbk:
After did a quick read on the code, it seems this is the way to go

Change `Pipeline.Write` interface to `Write(r io.Reader) ([]metatypes.Chunk, int, error) ` where `int` is the size of actual data.

And then modify these func to also return actual data
```go
(sop *SingleObjectPipeline) Write(r io.Reader) ([]metatypes.Chunk, error)
(asp *AsyncSplitterPipeline) Write(r io.Reader) ([]metatypes.Chunk, error)
```

`(sop *SingleObjectPipeline) Write(r io.Reader)` is very straightforward to modify.

For `(asp *AsyncSplitterPipeline) Write(r io.Reader) ([]metatypes.Chunk, error)`, we can modify this part

https://github.com/zero-os/0-stor/blob/1fed4cb32f51c31c5b9f99581eebdb0b8fe44ff9/client/datastor/pipeline/async_splitter.go#L157-L184

make it to also count the total data.

iwanbk:
what do you think @GlenDC ?

GlenDC:
I don't think I agree, but I do like the direction you're going. Will take a look at it later this week.

iwanbk:
>  Will take a look at it later this week.

[...]

PR #511, opened 2018-02-16
keep total size of data before processing in metastor.Metadata struct

zaibon:
closes: #500


client/client.go
```diff
@@ -438,3 +446,25 @@ func (ce closeErrors) Error() string {
...
+// readCounter is a io.Reader wrapper that counts the
+// total of bytes read by the underlying reader
+type readCounter struct {
+	r    io.Reader
+	size int64
+}
+
+// Read implement the io.Reader interface
+func (rc *readCounter) Read(p []byte) (n int, err error) {
+	if rc.r == nil {
```
GlenDC: You should be able to remove this check. `readCounter` is an internal struct, only used by ourselves. If `rc.r` equals `nil`, it means we have a bug in our code, and it deserves to crash.

client/client_test.go
```diff
@@ -190,8 +190,9 @@ func TestRoundTripGRPC(t *testing.T) {
 
 			// write data to the store
 			key := []byte("testkey")
-			_, err = c.Write(key, bytes.NewReader(data))
+			md, err := c.Write(key, bytes.NewReader(data))
 			require.NoError(t, err, "fail to write data to the store")
+			require.EqualValues(t, len(data), md.Size, "metadata.Size should be the same as the total size of the writtent file")
```
GlenDC: `written file")`*

and I would recommend to replace `the same as` with `equals`, it's a bit shorter and easier to read imho (but this is nitpicking, really, and not important as it's just a test error msg)

client/client_test.go
```diff
@@ -577,8 +579,11 @@ func testRepair(t *testing.T, config Config, repairErr error) {
 
 	meta, err := c.metastorClient.GetMetadata(key)
 	require.NoError(t, err)
+	require.EqualValues(t, len(data), meta.Size, "metadata.Size should be the same as the total size of the writtent file")
 	// store last-write epoch, so we can compare it later after repair
 	lastWriteEpoch := meta.LastWriteEpoch
+	// store data size, so we can compare it later after repair
+	Size := meta.Size
```
GlenDC: I guess the capitalized `Size` is a typo, and what you really want is `size`? For the name of the local variable here.

client/client_test.go
```diff
@@ -577,8 +579,11 @@ func testRepair(t *testing.T, config Config, repairErr error) {
 
 	meta, err := c.metastorClient.GetMetadata(key)
 	require.NoError(t, err)
+	require.EqualValues(t, len(data), meta.Size, "metadata.Size should be the same as the total size of the writtent file")
```
GlenDC: same typo here as earlier

client/client_test.go
```diff
@@ -611,6 +616,7 @@ func testRepair(t *testing.T, config Config, repairErr error) {
 	require.NoError(t, err)
 	require.NotNil(t, repairMeta)
 	require.True(t, repairMeta.LastWriteEpoch != 0 && repairMeta.LastWriteEpoch != lastWriteEpoch)
+	require.EqualValues(t, Size, repairMeta.Size)
```
GlenDC: Didn't know about the EqualValues method, very useful, I used to cast the types manually.

client/metastor/encoding/proto/metadata.proto
```diff
@@ -33,9 +33,12 @@ message Metadata {
     // and is chosen by the owner of this data.
     bytes key = 1;
 
-    // size in bytes represents the total size of all chunks,
+    // size in bytes represent the total size of the data before any processing
+    int64 size = 8 [(gogoproto.customname) = "TotalSize"];;
```
GlenDC: `;;` typo? Should be `;` I guess.

client/client.go
```diff
@@ -251,8 +251,15 @@ func (c *Client) Write(key []byte, r io.Reader) (*metatypes.Metadata, error) {
 		return nil, ErrNilKey // ensure a key is given
 	}
 
+	if r == nil {
```
GlenDC: Not sure if we really want to catch that, but then again we also catch the `nil`-key. I still didn't really figure out a good border between acceptable error and panic in cases like this.

In a way a panic seems reasonable, as having a `nil`-reader in cases like this is probably a bug. With the `key` that is debatable, as the key might be passed on as the result of some other operations.

But OK, as the write function is expected to be used with a lot of streamed data, I don't think this one check is going to hurt our performance, so I guess I'm OK with it. Even though it might be hard to really change something like this later to a panic again, as people might have started to rely on this error, for some weird reason.

But... I think you can leave it for now?

zaibon: yeah I guess panic is acceptable. But I had to add this return of error because of 
https://github.com/zero-os/0-stor/blob/1fed4cb32f51c31c5b9f99581eebdb0b8fe44ff9/client/client_test.go#L638-L641

Where we check that we return a proper error in case of nil reader.
What do you prefer to keep ?

GlenDC: I think I prefer it in the public client method function, I usually prefer to keep internal functions as nil-check free as much as possible, as you usually have enough control over the internal properties used.

PR #513, opened 2018-02-20
Fix async splitter.

iwanbk:
When using `Read` of io.Reader, it should consume the n > 0 bytes
returned before considering the error.

See https://golang.org/pkg/io/#Reader

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.


iwanbk:
This PR fixed https://github.com/gigforks/minio/issues/1

GlenDC:
Added the unit test myself. Thank you for this fix @iwanbk!

PR #519, opened 2018-02-23
List Keys

### LINKED ISSUE #502: List of objects in a dir
S3 and Minio has feature to list objects in a bucket (namespace in 0-stor term).
arguments:
- prefix : prefix of the object name
- max objects : max number of objects to show

Use case:
- browse the objects on a bucket  (using browser or another ftp-like client)
- before deleting the object

iwanbk:
Fixes #502

Add `ListKeys` API to `metastor.Client` object and `metastor.db.DB` interface, some notes:
- use `ListKeys` instead of `List` as name because we only want to list the keys
- `namespace` become field of `metastor.Client` object
-  still WIP because still need to add the API to the daemon. Creates this PR in my morning time instead of waiting for the daemon to be finished because i need to fly today. Will address the review and implement the daemon on monday morning


cmd/zstor/commands/file.go
```diff
@@ -205,6 +205,26 @@ var fileMetadataCmd = &cobra.Command{
...
+	Long:  "Print all files in this namespace",
+	Args:  cobra.ExactArgs(0),
+	RunE: func(_cmd *cobra.Command, args []string) error {
+		cl, err := getMetaClient()
+		if err != nil {
+			return err
+		}
+
+		return cl.ListKeys(func(key []byte) error {
+			fmt.Println(string(key))
```
zaibon: Since keys are []bytes, maybe we should add a flag to print it in hex in case the keys are pure binary?

iwanbk: added on 0372f864e6bb1f7a49c7d92e3e5efdf8a284cae7

Issue #516, opened 2018-02-21
Implement client.datastor.Client interface on top on 0-db


zaibon:
part of #515 

In order to be able to switch from zstordb to 0-db, we need to implement the [datastor.Client](https://godoc.org/github.com/zero-os/0-stor/client/datastor#Client) interface for 0-db

iwanbk:
How is the authentication mechanism?
I don't see jwt/IYO support in 0-db and see https://github.com/zero-os/0-db/issues/4 instead

iwanbk:
> How is the authentication mechanism?
> I don't see jwt/IYO support in 0-db and see zero-os/0-db#4 instead

Found the answer at https://github.com/zero-os/0-stor/issues/517#issue-299236129
"Once a namespace has been created on each shard, 0-stor client will receive a secret to authenticate against the 0-dbs and allow to access a namespace."

zaibon:
no indeed, JWT and IYO are out of the picture. 
So authentication is based on reservation and namespace.
An end user create a reservation for X amount of storages, a namespace with a quota of X is created on the 0-db but the 0-robot. The creation of the namespace will return a secret that needs to be used during the authentication to the 0-db.

So 0-stor client has a configuration for all the shards, a namespace and a secret for this namepsace.
At creation the 0-stor client send an AUTH command on all the shards to authentication the connection and select the namespace. From there the connection is secure and can be used freely

iwanbk:
One thing that confused me is this method
```go
   // Creates an object, using the given data.
    // The key of the object is generated by the server, and returned.
    CreateObject(data []byte) (key []byte, err error)
```

Based on the code i've read so far and manual test.
The object key is generated by server, with this format: `d:organization_0stor_namespace:sequence`.
`sequence` key will be incremented on each object write.(i.e. : if we upload `a.txt` twice, it will use two different keys)

See:
https://github.com/zero-os/0-stor/blob/1fed4cb32f51c31c5b9f99581eebdb0b8fe44ff9/server/db/memory/memory.go#L65-L69

and https://github.com/zero-os/0-stor/blob/1fed4cb32f51c31c5b9f99581eebdb0b8fe44ff9/server/db/badger/badger.go#L108-L114

To move forward, we have two options:
- add the key generation to 0-db
- put the key generation to client. I think this is the natural choice, most DBs i know let the user choose the key. But it seems we had a good reason to not do it?

zaibon:
[...]

PR #520, opened 2018-03-06
Zerodb

### LINKED ISSUE #515: Drop zstordb and use 0-db instead
After a discussions with @maxux and @despiegk regarding 0-db integration as backend for zstordb, it comes out that if 0-db already provide and network interface, zstordb doesn't really need to exists anymore since it's only job is to expose badger over grpc.

So instead of #514, what should happens is updating the client side of 0-stor to talk the protocol of 0-db.
Doing this will also allow us to close #504 #499 #488 #487 #352  cause all of these will be provided by 0-db.

There will be some change regarding how namespaces are created though.
With reservation in mind, the way how it will works is:
- a user reserve some storage space
- a 0-robot will received this reservation through the blockchain
- 0-robot contacts the 0-db running on top of the disk on the nodes and create a namespace on all the 0-db concerned. Creation of a namespace include the quota the 0-db can use on top of the disk
- the creation of a namespace return a secret that need to be used by the 0-stor client to authenticate for that namespace

iwanbk:
Closes #515 #516

Notes:
- `datastor.Client.ListObjectKeyIterator(ctx context.Context)` is still not implemented because 0-db still doesn't have the required feature (https://github.com/zero-os/0-db/issues/6). I don't see this func being used anywhere, so i send this PR instead of waiting for it to be implemented.
- codecov failed, it seems we need to remove `server` dir from codecov setting. Unfortunately i don't have privilege to do so.


README.md
```diff
@@ -96,7 +54,7 @@ All of these primitives are configurable and you can decide how your data will b
 
 ### etcd
 
-Other then a 0-stor server cluster, 0-stor clients also needs an [etcd](https://github.com/coreos/etcd) server cluster running to store it's metadata onto.
+Other then a 0-db server cluster, 0-stor clients also needs an [etcd](https://github.com/coreos/etcd) server cluster running to store it's metadata onto.
```
zaibon: can we also explain that the client exposes the metastor interface that can be implemented so we don't exactly must have an etcd to make it work.

client/config.go
```diff
@@ -54,9 +53,7 @@ func ReadConfig(path string) (*Config, error) {
 // storage clusters, as well as the entire read/write pipeline,
 // used to read and write data.
 type Config struct {
-	// IYO defines the IYO (itsyou.online) configuration,
-	// used for this 0-stor client.
-	IYO itsyouonline.Config `yaml:"iyo" json:"iyo"`
+	Password string `yaml:"password" json:"password"`
```
zaibon: Document this field

client/datastor/zerodb/client_test.go
```diff
@@ -0,0 +1,96 @@
...
+	obj, err := c.GetObject(key)
+	require.NoError(err)
+	require.Equal(key, obj.Key)
+	require.Equal(data, obj.Data)
+
+	nsObj, err := c.GetNamespace()
+	require.NoError(err)
+	require.NotNil(nsObj)
+	require.Equal(namespace, nsObj.Label)
+	// TODO : add number of entries check
```
zaibon: can we fix the todo now ?

client/datastor/zerodb/shard.go
```diff
@@ -14,19 +14,24 @@
...
-
-	"github.com/stretchr/testify/require"
+	"github.com/zero-os/0-stor/client/datastor"
 )
 
-func TestFreeSpace(t *testing.T) {
-	_, err := FreeSpace("")
-	require.Error(t, err)
+// Shard implements datastor.Shard for
+// 0-db clients, to make those clients work within a cluster of other GRPC clients.
```
zaibon: typo ? 'of other **GRPC** clients.' ?

iwanbk:
> some small comments
> And need to update Makefile to remove any trace of zstordb

fixed

GlenDC:
Please address that review @iwanbk in the dev-tft-launch though, when you can.

iwanbk:
> Please address that review @iwanbk in the dev-tft-launch though, when you can.

sure

PR #521, opened 2018-03-08
Add client.ReadRange API.

### LINKED ISSUE #499: offset and length option when reading object
To be used as minio backend, we need [client.Read](https://godoc.org/github.com/zero-os/0-stor/client#Client.Read) to have `start offset` and `length` parameter.

Amazon S3 API also allow this see `Range` header at https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

iwanbk:
Closes #499 

To read data in the given offset & length range.


client/client.go
```diff
@@ -273,6 +277,98 @@ func (c *Client) ReadWithMeta(meta metatypes.Metadata, w io.Writer) error {
...
+
+	if rw.length+int64(startIdx) < int64(endIdx) {
+		endIdx = startIdx + int(rw.length)
+	}
+
+	// call the underlying writer
+	// propagate the error if happens
+	n, err := rw.w.Write(p[startIdx:endIdx])
+	if err != nil {
+		return n, err
```
GlenDC: shouldn't we always compute  the length/offset, prior to checking for an error? Especially when the error is `io.EOF`.

iwanbk: > shouldn't we always compute the length/offset, prior to checking for an error?

Yep, you're right

iwanbk: fixed

client/datastor/pipeline/pipeline.go
```diff
@@ -57,6 +57,9 @@ type Pipeline interface {
 	// the details depend upon the specific implementation.
 	Delete(chunks []metatypes.Chunk) error
 
+	// ChunkSize returns maximal size of each chunk
```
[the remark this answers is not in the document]
iwanbk: fixed

PR #525, opened 2018-03-23
async_splitter.Read : deletes sent buffer entry.

iwanbk:
Deletes entry of buffer that already being sent to the Writer.

In current code, the buffer entry only getting deleted after the async splitter exited.

Part of OOM fix attempt reported at https://github.com/gigforks/minio/issues/30


GlenDC:
Thank you, very good catch!

PR #528, opened 2018-04-02
Use new zdb API for `direct` mode.

### LINKED ISSUE #527: Latest 0-db crashed
See https://github.com/rivine/0-db/issues/32, blocked by that issue.

It seems to need changes in 0-stor side.

iwanbk:
Fixes #527

Latest 0-db need the key to be empty.
This change is backward compatible with older 0-db.


iwanbk:
invalid issue, 0-db was buggy

